PR #83, opened 2018-08-07
Add --watch CLI option for watching and rebuilding subgraphs

### LINKED ISSUE #79: Watch Mode
Right now, after each change to a mapping, the developer has to complete the following steps: 
- Generate types to write your mappings
- Compile your subgraph and mappings
- Deploy your subgraph manifest to IPFS
- Deploy your subgraph to Graph Node
It would be much better if this could all be done in a single step during development and if that command could be executed on save.

fordN:
Resolves #79 

This PR adds a `--watch` option to the command line interface.  When passed into either `graph codegen` or `graph build` a watcher will be initialized on the working directory that detects file changes and triggers a restart of the respective process.  The watcher for `codegen` differs in that it ignores changes to typescript files (`*.ts`, `*.types.ts`).


graph-build.js
```diff
@@ -45,11 +48,56 @@ if (file === null || file === undefined) {
...
   verbosity: app.verbosity,
-}).compile()
+})
+
+// Watch working directory for file updates or additions, trigger compile (if watch argument specified)
+if (app.watch) {
+  subgraph_generator.logger.info('')
+  subgraph_generator.logger.info('%s %s', chalk.grey("Watching:"), process.cwd())
+  // Initialize watchers
+  let watcher = chokidar.watch('.', {
```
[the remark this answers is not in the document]
fordN: Great point, I've created a task to track the changes, #90.

fordN:
The watcher now parses the subgraph and watches the relevant files. 
- codegen: `ABI`'s, `subgraph manifest`
- build: `ABI`'s, `mappings`, `graphQL schema`, `subgraph manifest`

If a change is detected to the `subgraph manifest` it is parsed again and any newly added files will be watched.

examples/example-event-handler/package.json
```diff
@@ -3,10 +3,10 @@
   "scripts": {
     "codegen": "graph codegen --output-dir types/ subgraph.yaml",
     "build": "graph build subgraph.yaml",
-    "build-ipfs": "graph build --ipfs /ip4/127.0.0.1/tcp/5001 subgraph.yaml",
+    "build-ipfs": "graph build --watch --ipfs /ip4/127.0.0.1/tcp/5001 subgraph.yaml",
     "build-wast": "graph build -t wast subgraph.yaml"
   },
   "devDependencies": {
-    "graph-cli": "ssh://git@github.com/graphprotocol/graph-cli"
+    "graph-cli": "git+ssh://git@github.com/graphprotocol/graph-cli.git#ford/watch-manifest-files-only"
```
Jannis: Let's remember to remove this before we merge.

graph-codegen.js
```diff
@@ -24,5 +32,12 @@ if (file === null || file === undefined) {
 let generator = new TypeGenerator({
   subgraphManifest: file,
   outputDir: app.outputDir,
+  verbosity: app.verbosity
 })
-generator.generateTypes()
+
+// Watch working directory for file updates or additions, trigger type generation (if watch argument specified)
+if (app.watch) {
+  generator.watchAndCompile()
```
Jannis: This should be `watchAndGenerateTypes()`.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+      persistent: true,
+      ignoreInitial: true,
+      ignored: [
+        './dist',
+        './node_modules',
+        /(^|[\/\\])\../
+      ],
+      depth: 3,
+      atomic: 500
+    })
```
Jannis: Since we're only watching the files we're interested in now, some of the above parameters can probably be dropped, e.g. `ignored` and `depth`.

src/cli/compiler.js
```diff
@@ -51,7 +52,7 @@ class Compiler {
         : path.join(this.options.outputDir, 'subgraph.yaml')
 
     this.logger.info('')
-    this.logger.info(chalk.green('Completed'))
+    this.logger.info(chalk.green('Build Completed'))
```
Jannis: Let's make `Completed` lowercase.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
   }
 
+  getFileLocations() {
+    try {
+      let allFiles = []
+      let subgraph = this.loadSubgraph()
+
+      // Add all file paths specified in manifest
+      allFiles.push(subgraph.getIn(['schema', 'file']))
+      let mappingFiles = subgraph.get('dataSources').map(dataSource => {
```
Jannis: Drop the `let mappingFiles = ` here, since we're not using `mappingFiles` anywhere.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+          allFiles.push(dataSource.getIn(['mapping', 'file']))
+          dataSource.getIn(['mapping', 'abis']).map(abi => {
+            allFiles.push(abi.get('file'))
+          })
+      })
+
+      // Convert all paths to absolute reference
```
Jannis: What's "absolute reference"? How about `// Make paths absolute`? 😉

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
     }
   }
 
+  getFileLocations() {
```
Jannis: I'd make the purpose of the function even more clear: `getFilesToWatch`, then rename `allFiles` to `files`.

Then I'd also drop the `let allAbsolutePaths = ` and simply return the result of `files.map(...)`.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+      this.logger.fatal('Failed to parse subgraph file locations:', e)
+    }
+  }
+
+  watchAndCompile() {
+    let compiler = this
+    compiler.logger.info('')
+
+    // Initialize watcher
+    let watcher = chokidar.watch(path.resolve(this.sourceDir, this.options.subgraphManifest), {
```
Jannis: This path you shouldn't have to resolve / make absolute.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+    let fileLocations = this.getFileLocations()
+    watcher.add(fileLocations)
+
+    // Add event listeners
+    watcher
+      .on('ready', function() {
+        let files = watcher.getWatched()
+        compiler.logger.info('%s %j', chalk.grey("Watching:"), files)
+        compiler.compile()
+        watcher
```
Jannis: This `watcher.on('change', ...)` I'd move one level up to be side by side with `watcher.on('ready', ...)`.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+          .on('change', path => {
+            compiler.logger.info('%s %s', chalk.grey('File change detected: '), path)
```
Jannis: Across the CLI we're trying to display paths as short as possible. What you want instead of passing in `path` here is to use `this.displayPath(path)`.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+            if (path.endsWith('.yaml')) {
```
Jannis: This isn't good enough. Check for equality with the subgraph manifest file.

src/cli/compiler.js
```diff
@@ -66,6 +67,80 @@ class Compiler {
...
+              let newFiles = compiler.getFileLocations().filter(file => {
+                return fileLocations.indexOf(file) === -1
+              })
+              if (newFiles.length >= 1) {
+                this.add(newFiles)
```
Jannis: This is also not good enough. We can't add more and more files to the watcher, we should also remove files we're no longer watching.

So: Keep the watched files around e.g. via a `watchedFiles` array. Then, when there is a change to the manifest, obtain the `filesToWatchNext` (or something) and then either remove the old files and add the new files or diff them and only remove/add the ones that have changed.

src/cli/type-generator.js
```diff
@@ -114,4 +116,76 @@ module.exports = class TypeGenerator {
       this.logger.fatal('Failed to generate types for contract ABI:', e.stack)
     }
   }
+
```
Jannis: My comments about the compiler changes apply to the type generator as well.

Jannis:
@fordN There is one more thing missing in this PR: when there is e.g. a build error while watching, we should print the errors but keep watching. Because errors will happen and people will want to recover from them without having to restart watching.

src/cli/compiler.js
```diff
@@ -117,6 +117,7 @@ class Compiler {
           chalk.grey('File change detected: '),
           compiler.displayPath(path)
         )
+        compiler.logger.currentStep = 0
```
Jannis: Can we please do this in a place where it is _guaranteed_ to be reset whenever we compile — i.e. at the beginning of the `compile()` function?

src/cli/compiler.js
```diff
@@ -66,6 +67,102 @@ class Compiler {
...
+        compiler.logger.info(chalk.grey("Watching relevant manifest files"))
+        compiler.compile()
+      })
+      .on('change', path => {
+        compiler.logger.info('%s %s',
+          chalk.grey('File change detected: '),
+          compiler.displayPath(path)
+        )
+        compiler.logger.currentStep = 0
+        if (path === compiler.options.subgraphManifest) {
```
Jannis: Can you double-check also that these two have any chance of being equal? I'm assuming `path` will be absolute — is it guaranteed that `subgraphManifest` is? If not, we need to make it so that they can be compared.

src/cli/compiler.js
```diff
@@ -51,21 +61,143 @@ class Compiler {
...
+        .on('ready', function () {
+          compiler.logger.info('')
+          compiler.logger.info(chalk.grey('Watching relevant manifest files'))
+        })
+        .on('change', changedFile => {
+          compiler.logger.info('')
+          compiler.logger.info('%s %s',
+            chalk.grey('File change detected: '),
+            compiler.displayPath(changedFile)
+          )
```
fordN: I personally like having the file watcher updates as `logger.info, but it may make sense to hide in default setting .  Strong opinions?

src/cli/compiler.js
```diff
@@ -51,21 +61,143 @@ class Compiler {
...
+      })
+      return allAbsolutePaths
+    } catch(e) {
+      this.logger.fatal('Failed to parse subgraph file locations:', e)
+    }
+  }
+
+  watchAndCompile() {
+    let compiler = this
+    if (cluster.isMaster) {
```
Jannis: Is there any reason why we can't just catch exceptions thrown by `compile()` and keep watching? This cluster dance seems obscure but perhaps there are reasons for it?

src/cli/type-generator.js
```diff
@@ -85,6 +88,9 @@ module.exports = class TypeGenerator {
       return abis.map((abi, name) => this._generateTypesForABI(abi))
     } catch (e) {
       this.logger.fatal('Failed to generate types for contract ABIS:', e)
+    } finally {
+      this.logger.info('')
+      this.logger.info(chalk.green('Types generated'))
```
Jannis: This message breaks with the convention of using green message only for the top-level steps. Let's not include it at all.

src/cli/type-generator.js
```diff
@@ -108,10 +114,130 @@ module.exports = class TypeGenerator {
...
+            watcher.add(addedFiles)
+            watcher.unwatch(removedFiles)
+
+            if (addedFiles.length >= 1) {
+              let addedFilesDisplay = addedFiles.map(file => {
+                return generator.displayPath(file)
+              })
+              generator.logger.info(
+                '%s %j',
+                chalk.grey('Now watching: '),
```
Jannis: Several remarks here:

1. Let's use the debugging level for these kinds of messages.
2. Let's format the files nicely, the current output is not user-friendly. How about the following format?
```
Now watching:
  - abis/Foo.json
  - mappings/bar.ts
```

src/cli/type-generator.js
fordN: Ok. The reason I added it was because there was no confirmation of type generation completing when not in verbose mode.  Ok to have no logging in that situation? Or perhaps another way to include it, maybe as a different color?

src/cli/type-generator.js
fordN: sounds good :)

src/cli/compiler.js
fordN: I went with try catch originally, updating the `fatal` logger to no longer exit the process, but was getting a cascade of errors as it moved through the try catch blocks out to `compile()`.    These errors make sense, but it's a bit cumbersome and makes it less 'falling of the log easy' to pinpoint the issue.

fordN: Creating an exception constructor to throw the error and context message and then only logging at the top level.

src/cli/watcher.js
```diff
@@ -0,0 +1,74 @@
...
+    this.onCollectFiles = onCollectFiles
+    this.onError = onError
+  }
+
+  watch() {
+    // Collect files to watch
+    let files = this.onCollectFiles()
+
+    // Initialize watcher
+    this.watcher = chokidar.watch(files, {
```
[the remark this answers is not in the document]
Jannis: Oh, yeah! Good catch.

Jannis:
@fordN Comments should be addressed now.

src/cli/watcher.js
```diff
@@ -0,0 +1,84 @@
+const chokidar = require('chokidar')
+const path = require('path')
+
+module.exports = class Watcher {
+  constructor(options) {
+    const { onReady, onTrigger, onCollectFiles, onError, onFilesUpdated } = options
```
fordN: Is`onFilesUpdated` needed? It appears to be unused.

Jannis: Indeed it isn't used. I've removed it.

PR #85, opened 2018-08-09
Remove `Store.bind`

leodasvacas:
Essentially a revert of #60. Store is back to being just a namespace. Example is updated for our current preferred style, and the `dist` directory is added.

Also fixed small things that were off in `index.d.ts`.

Companion PR graphprotocol/graph-node#258


examples/example-event-handler/dist/ExampleSubgraph/abis/ExampleContract.json
```diff
@@ -0,0 +1,8 @@
+[
+  {
+    "anonymous": false,
+    "inputs": [{ "indexed": true, "name": "exampleParam", "type": "string" }],
+    "name": "ExampleEvent",
+    "type": "event"
+  }
+]
```
fordN: I don't think we want to include the ./dist/ folder in the repo.  Might be best to add ./dist to .gitignore.

leodasvacas: I think it's good, so when reviewing we can see the impact of a change on the example.

Jannis: I can live with either approach.

index.d.ts
```diff
@@ -215,7 +207,7 @@ declare class EthereumEvent {
   address: Address
   eventSignature: string
   blockHash: H256
-  params: Array<EthereumEventParam>
+  parameters: Array<EthereumEventParam>
```
Jannis: I'm ok with renaming this field but it seems unnecessary.

leodasvacas: The purpose is simply to match what's on `index.ts`

PR #89, opened 2018-08-15
Fix subgraph build issues caused by wrong types / missing coercions

Jannis:
This fixes build issues I've had when writing a new subgraph.


nenadjaja:
👍

Jannis:
@timmclean It accepts a byte array, its signature was just incorrectly saying `Address`.

timmclean:
Oh shoot read that backwards, yup all good

Issue #94, opened 2018-08-20
Add `graph remove` command


Jannis:

leodasvacas:
This also includes commands for removing and listing.

leodasvacas:
Since it is an admin-only thing, I don't think there's much value in providing list. Though remove is still to-do.

leodasvacas:
Done

PR #99, opened 2018-09-04
Add `graph deploy`

leodasvacas:
A convenience to compile and deploy a subgraph. No watch mode yet.

Example invocation:
`graph deploy subgraph.yaml --ipfs /ip4/127.0.0.1/tcp/5001 -n name --node http://localhost/`

Part of #94.


graph-deploy.js
```diff
@@ -0,0 +1,52 @@
...
+let requestUrl = new URL(app.node);
+if (!requestUrl.port) {
+    requestUrl.port = "8020"
+}
+
+compiler.compile()
+    .then(
+        function (ipfsHash) {
+            if (ipfsHash === undefined) {
+                console.log("compilation failed, not deploying")
```
Jannis: `compile` already logs something, so here we could just exit with an exit code of 1:
```js
process.exitCode = 1
return
```
There is also `process.exit(1)` but that [is not recommended](https://stackoverflow.com/questions/5266152/how-to-exit-in-node-js/37592669#37592669).

graph-deploy.js
```diff
@@ -0,0 +1,52 @@
...
+                return
+            }
+            requestUrl.pathname += `subgraph_deploy/${app.name}/${ipfsHash}`
```
Jannis: I'd use `path.join('subgraph_deploy', app.subgraphName, ipfsHash)` here.

BTW, it's `app.subgraphName`, not `app.name`.

graph-deploy.js
```diff
@@ -0,0 +1,52 @@
...
+            request.post(requestUrl, {
+                headers: {
+                    'Content-Type': 'application/json'
+                }
+            },
+                function (error, res, body) {
+                    if (error) {
+                        console.log("error sending json-rpc request " + error)
```
Jannis: Again: `console.error`. Also, since this a user facing error, I think formatting is important. I'd do:
```
console.error('Failed to deploy the subgraph:', error)
```

Jannis: Actually, we probably want to use a logger like in the compiler and type generator for this, to get consistent colors etc.

graph.js
```diff
@@ -8,4 +8,5 @@ app
   .command('build [file]', 'Compiles a subgraph and uploads it to IPFS', {
     isDefault: true,
   })
+  .command('deploy [file]', 'Deploys subgraph')
```
Jannis: This is a bit short. `"Deploys the subgraph to a graph node"` perhaps?

src/cli/args.js
```diff
@@ -0,0 +1,60 @@
...
+let app = require('commander')
+let path = require('path')
+let ipfsAPI = require('ipfs-api')
+
+let Compiler = require('./compiler')
+
+function addBuildCommand() {
+    app
+        .version('0.1.0')
+        .arguments('<cmd> [file]')
```
Jannis: These two lines above don't really belong in this function. Let's rename `args.js` to `app.js` and add a `createApp()` function that includes these two calls + the `--verbosity` option.

src/cli/args.js
```diff
@@ -0,0 +1,60 @@
...
+        ipfs,
+        subgraphManifest: file,
+        outputDir: app.outputDir,
+        outputFormat: app.outputFormat,
+        verbosity: app.verbosity,
+    })
+}
+
+exports.addBuildCommand = addBuildCommand
+exports.compilerFromArgs = compilerFromArgs
```
Jannis: Here you can do
```js
module.exports = {
  addBuildCommand,
  compilerFromArgs
}
```

leodasvacas:
@Jannis Review addressed and I'm now using a proper json-rpc lib as you suggested, that turned out great for error handling.

src/cli/app.js
```diff
@@ -0,0 +1,72 @@
...
+
+function addBuildCommand() {
+    app
+        .option(
+            '-o, --output-dir [path]',
+            'Output directory for build artifacts',
+            path.join(process.cwd(), 'dist')
+        )
+        .option('-t, --output-format [format]', 'Output format (wasm, wast)', 'wasm')
+        .option('i, --ipfs [addr]', 'IPFS node to use for uploading files')
```
Jannis: This one is my fault: But while we're at it, can we change `i` to `-i` here?

leodasvacas:
@Jannis done

Jannis:
@leodasvacas Haha, damn, can you run it again with semicolons disabled? Sorry!

leodasvacas:
@Jannis he, I should have suspected the diff size. I now ran with `--no-semi --single-quote --trailing-comma` which I believe is our preference.

Jannis:
@leodasvacas Correct, thank you!

PR #106, opened 2018-09-08
runtime: Add `store.get`

leodasvacas:


examples/example-event-handler/dist/ExampleSubgraph/ExampleSubgraph.ts
```diff
@@ -463,6 +464,41 @@ class Value {
   kind: ValueKind
   data: ValuePayload
 
+  toAddress(): Address {
+    assert(this.kind == ValueKind.BYTES, 'EthereumValue is not an address.')
```
Jannis: None of these are `EthereumValue`s, they are all `Value`s. Can you change the assertions?

examples/example-event-handler/dist/ExampleSubgraph/ExampleSubgraph.ts
```diff
@@ -539,6 +575,35 @@ class Value {
  * `Value` objects.
  */
 class Entity extends TypedMap<string, Value> {
+
+  getAddress(key: string): Address {
+    this.get(key).toAddress()
```
Jannis: The build outputs are outdated.

PR #109, opened 2018-09-08
Use namespaced package name

yanivtal:


package.json
```diff
@@ -1,9 +1,10 @@
 {
-  "name": "graph-cli",
-  "version": "0.2.4",
+  "name": "@graphprotocol/graph-cli",
+  "version": "0.2.5",
```
Jannis: The name change is fine.

In the future, we should only bump the version / publish releases from master though.

PR #110, opened 2018-09-08
Fix loading ABIs built by Truffle

Jannis:
These ABI files will be a map with a `contractName` and `abi` field,
rather than an array of type, function and event definitions.

If this is the case, we now load from the `abi` field, otherwise
we stick to the top level array.


Jannis:
@leodasvacas I've reworked the PR a little. Can you look again?

PR #111, opened 2018-09-08
getU256 and getI256 for Value

### LINKED ISSUE #107: Add `Entity::get` variants for different integer types

### LINKED ISSUE #108: Add type conversion from BigInt to different integer types

leodasvacas:
We use a host function exported in https://github.com/graphprotocol/graph-node/pull/343 to "downcast" the a BigInt to a U256 or I256.

Depends on #106 

Resolves #108 
Resolves #107


src/runtime/index.ts
```diff
@@ -539,6 +586,43 @@ class Value {
...
+  getString(key: string): string {
+    return this.get(key).toString()
+  }
+
+  getArray<T>(key: string): Array<T> {
+    return this.get(key).toArray()
+  }
+
+  getI256(key: string): I256 {
+    this.get(key).toI256()
```
fordN: it's not rust!  `return`

index.d.ts
```diff
@@ -20,6 +20,14 @@ declare namespace store {
...
    */
   function remove(entity: string, id: string): void
+
+  /**
+   * Fetches a previously created entity from the host store.
+   * 
+   * @param entity Name of the entity type.
+   * @param id Entity ID.
+   */
+  function get(entity: string, id: string): Entity
```
fordN: reminder: get() is not in /master yet on `graph-node`

PR #118, opened 2018-09-15
docs: Add info about `graph deploy`

html5cat:


README.md
```diff
@@ -66,6 +66,19 @@ An example of this can be found in the [Decentraland repository](https://github.
     yarn build      # Will drop the results in dist/
     yarn build-ipfs # Will also deploy to IPFS and output an IPFS hash
     ```
+9.  Deploy your subgraph to one of the hosted nodes. The following command builds and deploys the subgraph continuously as you are making changes to it:
+    ```sh
+    graph \
+       --watch \
+       --verbosity debug \
+       --node $GRAPH_NODE_MANAGEMENT \
```
Jannis: These environment variables are never mentioned anwyhere else. Let's add a `"deploy"` script in step 5, with specific values for the arguments and then mention that here.

README.md
```diff
@@ -66,6 +66,19 @@ An example of this can be found in the [Decentraland repository](https://github.
...
+       --ipfs $GRAPH_NODE_IPFS \
+       --subgraph-name my-subgraph \
+       --api-key $GRAPH_NODE_API_KEY \
```
Jannis: Let's not mention the `--api-key`, as it is a temporary solution. It's only needed when deploying to a Graph Node that requires authorization. The authorization we're currently will be removed as we move to the GNS etc.

html5cat:
@Jannis could you review updated code? Thanks!

html5cat:
@Jannis updated as per feedback – could you review and merge? Thanks!

README.md
```diff
@@ -66,6 +67,18 @@ An example of this can be found in the [Decentraland repository](https://github.
     yarn build      # Will drop the results in dist/
     yarn build-ipfs # Will also deploy to IPFS and output an IPFS hash
     ```
+9.  Deploy your subgraph to one of the hosted nodes. The following command builds and deploys the subgraph continuously as you are making changes to it:
```
Jannis: Hosted nodes are not a thing that's generally available. It would be better if this said "Deploy your subgraph to a Graph Node".

README.md
Jannis: The env vars are still there though. Let's instead use real values in this example.

PR #138, opened 2018-10-01
slight doc change to graph-cli

davekaj:
Just fixed some typos, and added some comments to make the feed back loop between graph cli and graph node main repo make more sense, for when new developers are looking at the repo

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,11 @@ project. It is recommended to install `graph-cli` as a local dependency
 via `package.json` and use `npm` scripts for code generation and
 building.
 
-An example of this can be found in the [Decentraland repository](https://github.com/graphprotocol/decentraland/).
+If you are just getting started with creating a subgraph, head to [getting started](https://github.com/graphprotocol/graph-node/blob/master/docs/getting-started.md). Eventually this guide will lead you back here.
+
+For clarity, an example of the setup below can be found in the [ENS subgraph repository](https://github.com/graphprotocol/ens-subgraph).
+
+### Steps 
 
 1.  Create a project for the subgraph with a `package.json` etc.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,17 +57,17 @@ An example of this can be found in the [Decentraland repository](https://github.
         "codegen": "graph generate-types subgraph.yaml",
         "build": "graph build subgraph.yaml",
         "build-ipfs": "graph build --ipfs /ip4/127.0.0.1/tcp/5001 subgraph.yaml",
-        "deploy": "graph deploy subgraph.yaml --watch --verbosity debug --node http://35.242.224.197:8020/ --ipfs /ip4/35.242.224.197/tcp/8030 --subgraph-name my-subgraph"
+        "deploy": "graph deploy subgraph.yaml --watch --verbosity debug --node http://<YOUR_IP_ADDRESS> --ipfs /ip4/<YOUR_IP_ADDRESS>/tcp/8030 --subgraph-name my-subgraph"
```
nenadjaja: Small fix: I would put IP_ADDRESS:PORT and explain below what that means:
- IP_ADDRESS is the IP address of your own node (for example, 35.242.224.197)
- PORT is the port number on which the node is running (for example. 8020)

Or something along those lines. So it's more clear :)

davekaj:
All changes have been addressed. However I noticed that we are still using `yarn build-ifps` here, and as noted in the other doc update I believe we want to use `yarn deploy` everywhere. so I will update that and then this can be merged

leodasvacas:
`deploy` will replace most uses of `build-ipfs`, but `build-ipfs` would still have its uses, unless we decide to deprecate it.

PR #140, opened 2018-10-03
Fix bugs in --watch mode for `codegen` and `build`

fordN:
This PR fixes the bugs in "watch mode" resolving #136.

---

### codegen bugs 
/src/type-generator.js
  - In `watchAndGenerateTypes` the `onTrigger` input variable did not match the input name.
  - `Watcher` was not being imported from `./watcher`

### build bugs 
/graph-build.js
  - The `-watch` arg was not being referenced correctly, so `compiler.watchAndCompile()` never ran

### DIFF
--- a/graph-build.js
+++ b/graph-build.js
@@ -1,5 +1,5 @@
 #!/usr/bin/env node
-
+let args = require('commander')
 let app = require('./src/app')
 
 app.initApp()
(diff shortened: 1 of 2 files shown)


Jannis (approved):
The argument is actually called `--watch`, not `-watch`. Great catches though! Tested and it all works.

PR #141, opened 2018-10-03
Update npm dependencies to the latest versions

fordN:


Jannis (changes requested):
Could we drop `package-lock.json` altogether and stick to `yarn.lock`? These two easily become out of sync. The updates themselves look good but if `yarn.lock` isn't updated, they won't take effect for anyone using Yarn.

Jannis:
Closing for now, we can redo it at some point.

PR #153, opened 2018-10-23
graph.js: Fix `graph deploy`

leodasvacas:


Jannis:
I'm merging this in right away, because this is a pretty severe bug.

PR #156, opened 2018-10-24
Support abi under compilerOutput

leodasvacas:
Makes abi loading more robust, doing better error checking and supporting abis under the compilerOutput key which is something that was found in the wild by @davekaj. Dave, could you check that this works for you?

My javascript style is probably crap, I'm sure @Jannis can give me some hints.

### DIFF
--- a/src/abi.js
+++ b/src/abi.js
@@ -268,8 +268,20 @@ module.exports = class ABI {
 
   static load(name, file) {
     let data = JSON.parse(fs.readFileSync(file))
-    return Array.isArray(data)
-      ? new ABI(name, file, immutable.fromJS(data))
-      : new ABI(name, file, immutable.fromJS(data.abi))
+
+    let abi = null
+    if (Array.isArray(data)) {
+      abi = data
+    } else if (data.abi !== undefined) {
+      abi = data.abi
+    } else if (data.compilerOutput !== undefined && data.compilerOutput.abi !== undefined) {
+      abi = data.compilerOutput.abi
+    }
+
+    if (!abi) {
+      throw Error(`Could not extract abi from file ${file}`)
+    }
+  
+    return new ABI(name, file, immutable.fromJS(abi))
   }
 }
--- a/src/abi.js
+++ b/src/abi.js
@@ -279,7 +279,7 @@ module.exports = class ABI {
     }
 
     if (abi === null || abi === undefined) {
-      throw Error(`Could not extract abi from ${file}`)
+      throw Error(`Could not extract ABI from ${file}`)
     }
   
     return new ABI(name, file, immutable.fromJS(abi))
--- a/src/abi.js
+++ b/src/abi.js
@@ -278,8 +278,8 @@ module.exports = class ABI {
       abi = data.compilerOutput.abi
     }
 
-    if (!abi) {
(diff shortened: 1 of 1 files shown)


src/abi.js
```diff
@@ -268,8 +268,20 @@ module.exports = class ABI {
...
+    let abi = null
+    if (Array.isArray(data)) {
+      abi = data
+    } else if (data.abi !== undefined) {
+      abi = data.abi
+    } else if (data.compilerOutput !== undefined && data.compilerOutput.abi !== undefined) {
+      abi = data.compilerOutput.abi
+    }
+
+    if (!abi) {
```
Jannis: JS being weird, this check would be better as `abi !== null && abi !== undefined`.

src/abi.js
```diff
@@ -268,8 +268,20 @@ module.exports = class ABI {
...
+      throw Error(`Could not extract abi from file ${file}`)
```
Jannis: s/abi/ABI/

Jannis: And remove the `file`, `${file}` by itself will make it pretty obvious that it's about a file.

davekaj:
It works. I'll test again when you fix jannis changes

leodasvacas:
@Jannis @davekaj review adressed

davekaj:
still works!

src/abi.js
```diff
@@ -268,8 +268,20 @@ module.exports = class ABI {
...
+    if (Array.isArray(data)) {
+      abi = data
+    } else if (data.abi !== undefined) {
+      abi = data.abi
+    } else if (data.compilerOutput !== undefined && data.compilerOutput.abi !== undefined) {
+      abi = data.compilerOutput.abi
+    }
+
+    if (abi === null || abi === undefined) {
+      throw Error(`Could not extract abi from ${file}`)
```
Jannis: Even though the field is called "abi" in JSON, let's spell it `ABI` for users because it's an intialism / acronym.

src/abi.js
```diff
@@ -268,8 +268,20 @@ module.exports = class ABI {
 
   static load(name, file) {
     let data = JSON.parse(fs.readFileSync(file))
-    return Array.isArray(data)
-      ? new ABI(name, file, immutable.fromJS(data))
-      : new ABI(name, file, immutable.fromJS(data.abi))
+
+    let abi = null
```
Jannis: You could write this as
```js
let abi = Array.isArray(data)
  ? data
  : data.abi !== undefined
    ? data.abi
    : data.compilerOutput !== undefined && data.compilerOutput.abi !== undefined
      ? abi.compilerOutput.abi
      : undefined

if (abi === undefined) {
  throw Error(`Could not extract ABI from ${file}`)
}
```

That's what I would do to avoid the awkward `let abi = null` assignment upfront. But not everyone likes ternary operators...

src/abi.js
leodasvacas: you mentioned that on the first review 🤦‍♂️

src/abi.js
leodasvacas: I'm not so used to that yet so I'll leave it as ifs.

Jannis: That's ok 👍

PR #163, opened 2018-11-02
Add auth command for saving auth tokens to the system's keychain

fordN:
New command added to save a node-authToken pair to the system's keychain for use by `deploy`. 
  - Update documentation to explain new `graph auth` command and to remind Linux users to ensure they have `libsecret` installed.
  - When deploying to a Graph node that requires auth now use `--auth` flag. If specified `deploy` will use the auth token that matches the node value provided. 
  - Update travis.yml 
      - Install `libsecret-1-dev`
      - CD into example project and run `yarn` before running scripts
      - Add `graph auth` script
  - Travis builds were getting false positives because of compiler failures returning with exit code = 0
      - Update  compiler to set exit code = 1 on failure


README.md
```diff
@@ -26,12 +27,19 @@ version of the subgraph to IPFS or a local directory for debugging.
...
 project. It is recommended to install `graph-cli` as a local dependency
 via `package.json` and use `npm` scripts for code generation and
-building.
+building. <br/>_
 
 If you are just getting started with creating a subgraph, head to [getting started](https://github.com/graphprotocol/graph-node/blob/master/docs/getting-started.md). Eventually this guide will lead you back here.
 
 For clarity, an example of the setup below can be found in the [ENS subgraph repository](https://github.com/graphprotocol/ens-subgraph).
 
+### On Linux
```
Jannis: Add empty line after section header.

README.md
```diff
@@ -26,12 +27,19 @@ version of the subgraph to IPFS or a local directory for debugging.
...
+`libsecret` is used for storing auth tokens, so you
+may need to install it before getting started. Use one of the following commands depending on your distribution:
```
Jannis: The newline here looks odd. Suggestion: balance the paragraph more evenly, like so
```
libsecret is used for storing auth tokens, so you may need to install it
before getting started. Use one of the following commands depending on
your distribution:
```

README.md
```diff
@@ -100,7 +108,8 @@ For clarity, an example of the setup below can be found in the [ENS subgraph rep
        --ipfs /ip4/127.0.0.1/tcp/5001 \
        --subgraph-name your-subgraph
     ```
-
+    _Note: if you are deploying to a hosted node use the `-auth` flag
```
Jannis: Let's rephrase this. Hosted nodes are not really a thing; we only set them up for hackathons so far. How about?

```
_Note: If the Graph Node you are deploying to requires authorization, make sure
to authorize with the node using e.g. 'graph auth http://127.0.0.1:8020 <access token>`
before deploying.
```

Let's not mention the `--auth` flag here, this isn't how it was supposed to work.

examples/example-event-handler/package.json
```diff
@@ -7,7 +7,7 @@
     "build-wast": "graph build -t wast subgraph.yaml"
   },
   "devDependencies": {
-    "@graphprotocol/graph-cli": "git+https://github.com/graphprotocol/graph-cli.git",
-    "@graphprotocol/graph-ts": "^0.0.1"
+    "@graphprotocol/graph-cli": "^0.3.2",
```
Jannis: Drop these changes, keep what's in master.

graph.js
```diff
@@ -59,6 +60,24 @@ function outputDeployConfig(cmd) {
   console.error('')
 }
 
+function outputAuthConfig(node, authToken) {
```
Jannis: We don't need this. If `graph auth` only has two inputs and if anything is wrong we should handle/report that in a more compact way.

Jannis: I take this back, but I'd like to request changes here:

* Change `Node:` to `Graph node:` to be consistent with the other outputs.
* Rename `authToken` to `accessToken`
* Change `AuthToken` to `Access token`
* Change `No auth token defined` to `Missing`
* Change `The auth token provided is too long` to `Access token is too long`
* Remove `AuthToken: Hidden` entirely

graph.js
```diff
@@ -133,17 +152,39 @@ app
     }
   })
 
+/**
+ * graph auth */
```
Jannis: Move the `*/` to the next line.

graph.js
```diff
@@ -133,17 +152,39 @@ app
...
+app
+  .command('auth [NODE] [AUTH_TOKEN]')
+  .description("Stores auth token on the system's keychain for use by deploy")
+  .action((node, authToken) => {
+    if (
+      authToken === undefined ||
+      node === undefined ||
+      authToken > 200
```
Jannis: This should be `authToken.length`, I think?

Please rename `AUTH_TOKEN` and `authToken` to `ACCESS_TOKEN` and `accessToken`.

Also: `Stores the auth token on the system's keychain for use by deploy` is not the most user-friendly message. How about this?
```js
.description("Sets the access token to use when deploying to a Graph node")
```

graph.js
```diff
@@ -133,17 +152,39 @@ app
...
+    ) {
+      outputAuthConfig(node, authToken)
+      console.error('--')
+      console.error('For more information run this command with --help')
+      process.exitCode = 1
+      return
+    }
+    keytar.setPassword('subgraph-hosting', node, authToken).then(() => {
```
Jannis: `subgraph-hosting` isn't great. How about `graphprotocol-auth`?

graph.js
```diff
@@ -133,17 +152,39 @@ app
...
+      console.log("Auth token saved on keychain for Graph node:", node)
```
Jannis: Make this `"Access token saved for Graph node:", node)`.

graph.js
```diff
@@ -133,17 +152,39 @@ app
...
+    }).catch((err) => {
+      console.error('Failed to save auth token to the keychain: ', err)
```
Jannis: `'Failed to save access token:'`

graph.js
```diff
@@ -133,17 +152,39 @@ app
...
  */
 app
   .command('deploy [SUBGRAPH_MANIFEST]')
   .description('Deploys the subgraph to a graph node')
-  .option(
-    '-k, --api-key <KEY>',
-    'Graph API key authorized to deploy to the subgraph name'
-  )
   .option('-g, --node <URL>[:PORT]', 'Graph node to deploy to')
+  .option('-a, --auth', 'Use auth token for deployment')
```
Jannis: That's not how it's supposed to work. First of all, let's not introduce a short `-a` flag for now. The long flag should be `--access-token <ACCESS_TOKEN>`, as a way to override the (potentially) saved access token for a node.

graph.js
```diff
@@ -183,6 +224,20 @@ app
 
     let logger = new Logger(0, { verbosity: getVerbosity(app) })
 
+    if (cmd.auth) {
+      keytar.getPassword('subgraph-hosting', cmd.node).then((authToken) => {
```
Jannis: Rework this logic to take both the stored access token and then one optionally passed in with `--access-token` into account.

Jannis:
@fordN Reviewing this, I realized we didn't communicate that Tim used the term `access token` in graph-node already. Let's settle on that terminology and call it 'access token' everywhere.

README.md
```diff
@@ -26,12 +27,21 @@ version of the subgraph to IPFS or a local directory for debugging.
 Subgraphs for The Graph are set up just like any regular TypeScript
 project. It is recommended to install `graph-cli` as a local dependency
 via `package.json` and use `npm` scripts for code generation and
-building.
+building. <br/>_
```
Jannis: Let's remove this and just stick to regular paragraphs.

.travis.yml
```diff
@@ -4,9 +4,12 @@ node_js:
...
 cache: yarn
+before_install:
+  - sudo apt-get install libsecret-1-dev
 script:
   - yarn test
   - cd examples/example-subgraph
   - yarn
   - ../../graph.js codegen --output-dir src/types --debug
   - ../../graph.js build --debug
+  - ../../graph.js auth node authtoken
```
Jannis: This still says "authtoken", where it should maybe say "example-access-token".

README.md
```diff
@@ -5,10 +5,11 @@
 
 The Graph command line interface.
 
-As of today, the command line interface consists of two commands:
+As of today, the command line interface consists of three commands:
 
 - `graph codegen` — generates TypeScript code for smart contract ABIs used in subgraphs.
 - `graph build` — compiles subgraphs to WebAssembly and deploys them to IPFS.
+- `graph auth` — saves access token to system's keychain.
```
Jannis: Change to `saves access tokens for Graph Nodes to the system's keychain`.

README.md
```diff
@@ -100,6 +110,9 @@ For clarity, an example of the setup below can be found in the [ENS subgraph rep
        --ipfs /ip4/127.0.0.1/tcp/5001 \
        --subgraph-name your-subgraph
     ```
+    _Note: If the Graph Node you are deploying to requires authorization,
+    make sure to authorize with the node using e.g. `graph auth http://127.0.0.`:8020 <ACCESS TOKEN>`
```
Jannis: Change to `ACCESS_TOKEN`

graph.js
```diff
@@ -59,13 +60,29 @@ function outputDeployConfig(cmd) {
...
+}
+
 /**
  * Global app configuration and options
  */
 app
   .version(module.exports.version)
   .option(
-    '--verbosity <info|verbose|debug>',
+    '--verbosity <INFO|VERBOSE|DEBUG>',
```
Jannis: Erm, no. These are real values, not placeholders.

graph.js
```diff
@@ -108,14 +125,14 @@ app
...
+  .option('-i, --ipfs <ADDR>', 'IPFS node to use for uploading files')
   .option('-n, --subgraph-name <NAME>', 'Subgraph name')
   .option(
-    '-o, --output-dir <path>',
+    '-o, --output-dir <PATH>',
     'Output directory for build results',
     path.resolve(process.cwd(), 'dist')
   )
-  .option('-t, --output-format <wasm|wast>', 'Output format (wasm, wast)', 'wasm')
+  .option('-t, --output-format <WASM|WAST>', 'Output format (wasm, wast)', 'wasm')
```
Jannis: Same here: these are real values, not placeholders.

graph.js
```diff
@@ -137,21 +154,44 @@ app
...
+      accessToken.length > 200
+    ) {
+      outputAuthConfig(node, accessToken)
+      console.error('--')
+      console.error('For more information run this command with --help')
+      process.exitCode = 1
+      return
+    }
+    keytar.setPassword('graphprotocol-auth', node, accessToken).then(() => {
+      console.log("Access token saved for Graph node:", node)
```
Jannis: Did you run `prettier` against this? I see a mix of `"` and `'` in this file. Single quotes are the way to go.

graph.js
```diff
@@ -137,21 +154,44 @@ app
...
+    }).catch((err) => {
+      console.error('Failed to save access token: ', err)
```
Jannis: Remove the space after `token:`, otherwise we'll log two spaces in that place.

graph.js
```diff
@@ -181,12 +221,21 @@ app
...
 
+    if(cmd.accessToken !== undefined) {
+      client.options.headers = { Authorization: 'Bearer ' + cmd.accessToken }
+    } else {
+      keytar.getPassword('graphprotocol-auth', cmd.node).then((accessToken) => {
+        if (accessToken !== null) {
+          client.options.headers = {Authorization: 'Bearer ' + accessToken}
+        }
+      }).catch((err) => {
+        logger.fatal("Error fetching access token: ", err)
```
Jannis: Double quotes. Remove the space after `token:`.

src/logger.js
```diff
@@ -161,6 +161,7 @@ module.exports = class Logger {
       }
     } else {
       this.fatal('Failed to compile subgraph', e)
+      process.exitCode = 1
```
Jannis: Let's leave it to the caller to decide whether failing to compile a subgraph is reason enough to exit with an error code. This doesn't belong in the logger.

graph.js
```diff
@@ -137,21 +154,44 @@ app
     }
   })
 
+/**
+ * graph auth
+ */
+app
+  .command('auth [NODE] [AUTH_TOKEN]')
```
Jannis: This still says `AUTH_TOKEN`.

graph.js
```diff
@@ -181,12 +221,21 @@ app
...
-    if (cmd.apiKey !== undefined) {
-      client.options.headers = { Authorization: 'Bearer ' + cmd.apiKey }
-    }
 
     let logger = new Logger(0, { verbosity: getVerbosity(app) })
 
+    if(cmd.accessToken !== undefined) {
+      client.options.headers = { Authorization: 'Bearer ' + cmd.accessToken }
+    } else {
+      keytar.getPassword('graphprotocol-auth', cmd.node).then((accessToken) => {
```
Jannis: `cmd.node` isn't going to work. All it takes is `http://some-ip:8020/` to be passed in instead of `http://some-ip:8020` and it won't work.

Let's do this:
```
let node = normalizeNodeUrl(cmd.node)
```
before we pass it in here, and also when we store the access token in the first place.
What normalizeNodeUrl should do is return the same URL regardless of whether there was a leading slash, leading or trailing whitespace etc. Ideally use https://nodejs.org/api/url.html for this.

graph.js
```diff
@@ -166,7 +166,7 @@ app
     let logger = new Logger(0, { verbosity: getVerbosity(app) })
 
     if (accessToken === undefined || nodeUrl === undefined || accessToken.length > 200) {
-      console.error('Failed to set access token')
+      console.error('Information missing to set the access token')
```
fordN: If accessToken is too long this error message may not help quickly identify the issue.

Jannis: Ah, yeah, `missing` is not the best message here. I'll update it.

Jannis: Done

Jannis:
@fordN I tried a few things to get D-Bus + keytar to work in Travis but I couldn't get it to work. I don't think it's worth it for this little command, so I've removed the `graph auth` command from Travis.

fordN:
@Jannis I agree, we can leave it off the Travis build. Odd that this issue didn't come up earlier.

PR #170, opened 2018-11-14
Test building code that uses BigInt math in the example subgraph

Jannis:
This makes sure we at least test building the changes added in `graph-ts` and `graph-node`:

* https://github.com/graphprotocol/graph-ts/pull/20
* https://github.com/graphprotocol/graph-node/pull/574


examples/example-subgraph/dist/subgraph.yaml
```diff
@@ -16,7 +16,7 @@ dataSources:
         - ExampleEntity
       abis:
         - name: ExampleContract
-          file: ExampleSubgraph/abis/ExampleContract.json
+          file: abis/ExampleContract.json
       eventHandlers:
         - event: ExampleEvent(string)
           handler: handleExampleEvent
```
fordN: The handler name does not match the function name in mapping.ts.

Jannis: Good catch — fixed!

Jannis:
I'm going to be a naughty boy and merge this, so we can cut new releases for graph-ts, graph-cli and graph-node.

PR #171, opened 2018-11-19
Travis: Run `graph auth` as part of the test commands

Jannis:
This requires a 'login' keyring to be created upfront using gnome-keyring.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,10 +5,17 @@ node_js:
   - "10"
 cache: yarn
 before_install:
-  - sudo apt-get install libsecret-1-dev
+  - sudo apt-get install libsecret-1-dev dbus gnome-keyring python-keyring python-gnomekeyring
+before_script:
+  # Create a new 'login' keyring; this appearas to be necessary for gnome-keyring to auto-launch
+  # properly when needed
+  - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -5,10 +5,17 @@ node_js:
...
+  - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
 script:
   - yarn test
   - cd examples/example-subgraph
   - yarn
   - ../../graph.js codegen --output-dir src/types --debug
   - ../../graph.js build --debug
+  # Run `graph auth` inside `dbus-launch` to be able to access gnome-keyring
+  # secrets via keytar
+  - dbus-launch bash -c '../../graph.js auth http://some-node-ip.org test-access-token'
```
fordN: It would be nice to mention the gnome-keyring setup and dbus-launch in the README for linux users to reference if they run into the same issue. 
Will these steps be necessary for a specific set of linux distributions?

Jannis: Linux desktop users will already be running a D-Bus session bus, so they don't need special instructions. This tweak also only worked for Ubuntu Trusty, not Precise. It would be nice to have clear instructions for this but right now the situation looks pretty messy to me, so I'd rather not add extra confusion.

fordN:
LGTM 👍

fordN (approved):
👍

Issue #148, opened 2018-10-10
Subgraph validation and normalization


Jannis:
# Validation

- [ ] Validate subgraph manifest using a GraphQL schema
- [ ] Validate the subgraph's GraphQL schema
- [ ] Validate ABIs in the subgraph
- [ ] Validate that events are present in ABIs

# Normalization

- [ ] Normalize subgraph manifest (deterministic field and list item order etc.)
- [ ] Normalize event signatures (e.g. `uint` -> `uint256`, no parameter names, no spaces)
- [ ] Normalize mappings (e.g. format with prettier)

leodasvacas:
The normalization plan is under the assumption that we're going to read a yaml, normalize it, output another yaml and hash that. One thing that would complicate this is supporting subgraph manifests in json format, which would require a step converting the json to yaml. But the other way around of normalizing yaml to json seems simpler, and json also has the advantage of being easier to put in a canonical shape since it doesn't care about whitespace.

The point is that if we'll support both yaml and json manifests, we should consider having json be the canonical format because it's simpler.

Jannis:
Agreed! The above doesn't make any assumptions like that though — only the current PR does. But YAML is what we use right now, so it seems appropriate. Plus the logic there should also work for JSON.

leodasvacas:
@Jannis wanna check off all that got done here?

Jannis:
@leodasvacas Done, I would've forgotten :)

PR #174, opened 2018-11-21
Add more validations for subgraphs

Jannis:
This PR adds the following validations, potentially saving hours of debugging:

* Validate ABIs in the subgraph
* Validate that events are present in ABIs (without normalization for now)
* Validate that all field types in the schema are valid
    - This includes suggestions for common problems, like `Unknown type 'Address'. Did you mean 'Bytes'?`.
* Validate that all entity types have an id field of type ID!
* Validate that all contract addresses in the manifest are valid

Below are the error messages for these validations:

## Invalid ABIs

```
$ cat examples/example-subgraph/abis/ExampleContract.json.foo
"Hello"

$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in examples/example-subgraph/subgraph.yaml:

    Path: dataSources > 0 > mapping > abis > 0 > file
    No valid ABI in file /Users/jannis/Work/graphprotocol/graph-cli/examples/example-subgraph/abis/ExampleContract.json.foo
```

## Events not found in ABI

```
$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in examples/example-subgraph/subgraph.yaml:

    Path: dataSources > 0 > eventHandlers > 0
    Event with signature ExampleEvent(string,uint256) not present in ABI "ExampleContract". Candidates: ExampleEvent(string)
```

## Invalid entity field types

```
$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in /Users/jannis/Work/graphprotocol/graph-cli/examples/example-subgraph/schema.graphql:

  Type 'ExampleEntity', field 'optionalBoolean': Unknown type 'bool'. Did you mean 'Boolean'?

  Type 'OtherEntity', field 'owner': Unknown type 'Address'. Did you mean 'Bytes'?
```

## Missing `id: ID!` fields on entities

```
$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in /Users/jannis/Work/graphprotocol/graph-cli/examples/example-subgraph/schema.graphql:

  Entity type 'OtherEntity': Missing field: id: ID!
```

## Invalid contract addresses

Address too long:
```
$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in examples/example-subgraph/subgraph.yaml:

    Path: dataSources > 0 > source > address
    Contract address must have length 40 (or 42 if prefixed with 0x) but has length 41: 22843e74c59580b3eaf6c233fa67d8b7c561a8359
```

Address not hexadecimal:
```
$ graph codegen examples/example-subgraph/subgraph.yaml
Failed to generate types: Failed to load subgraph: Error in examples/example-subgraph/subgraph.yaml:

    Path: dataSources > 0 > source > address
    Contract address is not a hexadecimal string: 0x22843e74c59580b3eaf6c233fa67d8b7c561a83X
```


Jannis:
This ticks most currently tracked validation ideas in https://github.com/graphprotocol/graph-cli/issues/148.

leodasvacas:
More validation is always great! Since you're touching this code, I have a request, could we remove (or somehow downgrade to a non-fatal warning) the "Unexpected key in map" validation? It makes adding new fields to the manifest a breaking change, which is unfourtunate.

Jannis:
@leodasvacas Shouldn't we rather add those fields to the schema that we use for the validation?

leodasvacas:
Yes, but I'd prefer that being able to add the fields is not tied to upgrading the CLI.

Another thing, right now all of the fields in `manifest-schema.graphql` are mandatory, even if they are nullable. I tried adding a nullable field and got " No value provided" if it wasn't there. Probably we should allow nullable fields to not be present.

Jannis:
I think it's a good thing that the validation in CLI needs to be updated first. We should stick to these structure validations.

The other thing would be a bug that we can fix easily.

leodasvacas:
> I think it's a good thing that the validation in CLI needs to be updated first. We should stick to these structure validations.

For now this is ok. But once we start promising backwards-compatibility and versioning becomes an issue both for code maintenence and for users that need to upgrade, it's unlikely that we'll consider this validation to be worth the trouble.

> The other thing would be a bug that we can fix easily.

Cool, I'll give it a shot after this lands.

src/subgraph.js
```diff
@@ -17,42 +31,195 @@ module.exports = class Subgraph {
...
+              path,
+              message:
+                `Contract address must have length 40 (or 42 if prefixed with 0x) ` +
+                `but has length ${address.length}: ${address}`,
+            })
+          )
+        }
+
+        // Validate whether the address is a hex string
+        let pattern = /^(0x[0-9a-fA-F]{40}|[0-9a-fA-F]{40})$/
```
leodasvacas: I'm not great with regex, but maybe this be better written as `(0x)?[0-9a-fA-F]{40}`.

This check is a more precise version of the one above it, we could remove the first one, and adjust the error message to mention the accepted formats.

leoyvens (changes requested):
The validations, type suggestions and tests are all pure ❤️. I just left a small comment. Also the example subgraph is not passing one of the validations.

The tests for error message output remind of how rustc does that, each test has a `.stderr` file beside it with the expected output, and when many tests change at once there is a tool to override the `.stderr` files with the new output. Not that we have to be this fancy right now, just to say that this is a very important kind of test.

src/subgraph.js
Jannis: Haha, you are right about the regular expression. 👍 Let's combine the two error messages into one like you're suggesting.

Jannis:
I've removed the `.exit` files as per @leodasvacas's request. Instead, `{ exitCode: 1 }` is passed to `testCli()`.

Other changes:

* Test stdout and stderr for _equality_, rather than just matching against the expected output. This required printing paths relative to the working directory in errors, which makes sense anyway.
* Group all schema errors by entity: https://github.com/graphprotocol/graph-cli/blob/966b79bb2a6192f4e22ff08f4f312d80a9c04e3b/tests/cli/validation/invalid-entity-field-types.stdout#L1-L15

I think that's it. I'll look into why the tests are failing still.

leoyvens (approved):

PR #177, opened 2018-11-27
Continue without access token if keytar fails

timmclean:
Warns if `getPassword` fails, but keeps going without an access token. Useful on Linux in headless setups.
Also added some platform-specific messages to explain to users what graph-cli is trying to do.

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -181,13 +181,17 @@ app
       logger.status('Access token set for Graph node:', node)
     } catch (e) {
       if (process.platform === 'win32') {
-        logger.error(`Error storing Graph access token in Windows Credential Vault:`, e)
+        logger.error(`Error storing access token in Windows Credential Vault:`, e)
       } else if (process.platform === 'darwin') {
-        logger.error(`Error storing Graph access token in macOS Keychain:`, e)
+        logger.error(`Error storing access token in macOS Keychain:`, e)
       } else if (process.platform === 'linux') {
-        logger.error(`Error storing Graph access token with libsecret (usually gnome-keyring or ksecretservice):`, e)
+        logger.error(
+          `Error storing access token with libsecret ` +
+          `(usually gnome-keyring or ksecretservice):`,
+          e
+        )
       } else {
-        logger.error(`Error storing Graph access token in OS secret storage service:`, e)
+        logger.error(`Error storing access token in OS secret storage service:`, e)
       }
       process.exitCode = 1
     }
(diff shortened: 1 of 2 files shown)


graph.js
```diff
@@ -180,7 +180,15 @@ app
       await keytar.setPassword('graphprotocol-auth', node, accessToken)
       logger.status('Access token set for Graph node:', node)
     } catch (e) {
-      logger.error('Failed to set access token:', e)
+      if (process.platform === 'win32') {
+        logger.error(`Error storing Graph access token in Windows Credential Vault:`, e)
```
Jannis: I'd just make these `access token` instead of `Graph access token`.

graph.js
```diff
@@ -241,13 +249,20 @@ app
         let node = normalizeNodeUrl(cmd.node)
         accessToken = await keytar.getPassword('graphprotocol-auth', node)
       } catch (e) {
-        logger.error(`Failed to fetch access token:`, e)
-        process.exitCode = 1
-        return
+        if (process.platform === 'win32') {
+          logger.errorWarning(`Could not get Graph access token from Windows Credential Vault: `, e)
```
Jannis: `Graph access token` -> `access token`

graph.js
```diff
@@ -180,7 +180,15 @@ app
       await keytar.setPassword('graphprotocol-auth', node, accessToken)
       logger.status('Access token set for Graph node:', node)
     } catch (e) {
-      logger.error('Failed to set access token:', e)
+      if (process.platform === 'win32') {
+        logger.error(`Error storing Graph access token in Windows Credential Vault:`, e)
+      } else if (process.platform === 'darwin') {
+        logger.error(`Error storing Graph access token in macOS Keychain:`, e)
+      } else if (process.platform === 'linux') {
+        logger.error(`Error storing Graph access token with libsecret (usually gnome-keyring or ksecretservice):`, e)
```
Jannis: This line is a little too long. You can break the string up with `+` or by using backslashes to spread it across multiple lines.

graph.js
```diff
@@ -241,13 +249,20 @@ app
...
       } catch (e) {
-        logger.error(`Failed to fetch access token:`, e)
-        process.exitCode = 1
-        return
+        if (process.platform === 'win32') {
+          logger.errorWarning(`Could not get Graph access token from Windows Credential Vault: `, e)
+        } else if (process.platform === 'darwin') {
+          logger.errorWarning(`Could not get Graph access token from macOS Keychain: `, e)
+        } else if (process.platform === 'linux') {
+          logger.errorWarning(`Could not get Graph access token from libsecret (usually gnome-keyring or ksecretservice): `, e)
```
Jannis: Long line, see above.

Jannis (changes requested):

Jannis (approved):

PR #179, opened 2018-11-28
compiler.js: Append `global.ts` to mappings

leoyvens:
<strike>In the lack of a more inventive approach for https://github.com/graphprotocol/graph-ts/issues/7, we just append `global.ts` to put things in the global namespace.</strike>

<strike>This requires creating a temporary file with the appended contents. We can't put it in `/tmp/` because the mapping may be using relative imports, and copying the whole directory to `/tmp/` would be slow. So this creates a temporary hidden file in the same directory as the input file. Not super clean but the best I could come up with.</strike>

@Jannis, who has become a master in the black arts of invoking `asc`, found out that all we needed to cleanly do this is to pass both the mapping and `global.ts` as input files to asc.

We this is released, we should give everyone a heads up to remove the import from their mappings.

Depends on https://github.com/graphprotocol/graph-cli/pull/179 (also needed for tests to pass).


Jannis (dismissed):
Yes!

leodasvacas:
@Jannis That was a ninja approve! Can we do https://github.com/graphprotocol/graph-ts/pull/26 first?

Jannis:
@leodasvacas Done!

Jannis (approved):

PR #180, opened 2018-12-02
Add Entity.load(id) and entity.save() methods

Jannis:
This introduces the following changes:

1. The generated `SomeEntity` constructor now requires an ID to be provided (e.g. `new Collateral('some-id')`). This is a breaking change that guarantees there is always an ID.

2. A new static method `SomeEntity.load(id: string): SomeEntity | null` is generated for entity types, preventing developers from having to call `store.get('SomeEntity', id)` and then having to cast the return type from `Entity | null` to `SomeEntity | null`, which is not obvious.

3. A new `someEntity.save()` method is generated for entity types. It is a simple wrapper around `store.set('SomeEntity', id, entity)`. Makes life a little easier for developers and also asserts that there is a value for `id`.


leodasvacas:
This a much nicer API! What about renaming `getById` to something simpler like `get` or `find`?

We should be careful with generating constructors of subclasses (like we do now with `Entity`) because they won't call the super constructor. And calling `super()` doesn't in AS right now.

Jannis:
@leodasvacas Agree about being careful about constructors. 

My first thought was `get` but then that conflicts with the `get` method inherited from `TypedMap`. `find` could work but I'd like to keep the available for more powerful queries later. How about `findOne`?

Jannis:
Other idea: `SomeEntity.getOne(id: string)`

Zerim:
If we're going w/ an ORM-style API, I think of `fetch` or `load` as being the inverse operations of `save`.

Jannis:
@leodasvacas @Zerim Changed it to `SomeEntity.load(id)`. Remember that this still uses `store.get()` and `store.set()` under the hood, so these two methods are just convenience wrappers.

examples/example-subgraph/src/types/schema.ts
```diff
@@ -2,12 +2,50 @@ import {
...
+    this.set("id", Value.fromString(id));
+    return this;
+  }
+
+  save(): void {
+    let id = this.get("id");
+    assert(id !== null, "Cannot save ExampleEntity entity without an ID");
+    assert(
+      id.kind == ValueKind.STRING,
+      "Cannot save ExampleEntity entity with non-string ID"
```
davekaj: I think we could add a helper line to the text. Like: "Consider using toHex() to change the Ethereum type to a string"

davekay100 (commented):
We can also update the `subgraph.yaml` to have `OtherEntity` added, and put `0x` in the address. 

Also, I have been building subgraphs without the `/src` folder. The top level I have is always `/mappings` and `/types` side by side with `/abis` and everything else. Here we have `mappings` and `types` inside `src`. 

I think we should pick one way to do it for all subgraphs. But I think the way you did it here is better, so I will start updating the subgraphs to reflect that

davekaj:
Otherwise it looks good to me. I would have liked to have tested the subgraph, but I guess that contract doesn't exist anywhere in the real world .

Jannis:
@davekaj I've made the changes you've asked for.

davekay100 (approved):
approved

PR #184, opened 2018-12-09
Make `graph auth` and `--access-token` useful for IPFS as well

Jannis:
Whenever users deploy a Graph Node in public, they will want or need to protect Graph Node and their IPFS node behind authentication.

This PR makes `graph auth` able to store access tokens for IPFS as well, like with
```sh
graph auth /ip4/127.0.0.1/tcp/80 some-token
```

It also makes `graph build --ipfs ...` and `graph deploy` lookup the access token for both Graph Node and IPFS from the keychain. Note that when providing the access token with `--access-token`, the same access token is used for IPFS and Graph Node.

This PR also fixes `graph remove` to look up access tokens from the keychain or take it in with `--access-token`.

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -34,8 +34,53 @@ function createCompiler(app, cmd, subgraphManifest) {
   })
 }
 
-function normalizeNodeUrl(node) {
-  return new URL(node).toString()
+// Helper function to obtain the access token for the target node
+async function getAccessToken(cmd, url, logger) {
+  // Determine the access token to use, if any:
+  // - First try using --access-token, if provided
+  // - Then see if we have an access token set for the Graph node
+  if (cmd.accessToken !== undefined) {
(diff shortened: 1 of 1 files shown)


graph.js
```diff
@@ -34,8 +46,53 @@ function createCompiler(app, cmd, subgraphManifest) {
...
+      }
+      logger.status(`Continuing without an access token`)
+    }
+    return undefined
+  }
+}
+
+function normalizeUrl(url) {
+  try {
+    return new URL(url).toString()
```
timmclean: Is it OK that this won't strip a trailing slash?

timmclean (changes requested):

graph.js
Jannis: Kind of, maybe. We do need better URL normalization here.

Issue #169, opened 2018-11-14
Generated attribute setter mishandles false boolean values


fordN:
Boolean values will have to be treated differently than other types in the generated attribute setters because boolean false is equal to null in compiled wasm.  
See https://github.com/graphprotocol/graph-node/issues/571 for information.

Jannis:
I've investigated this briefly and wrote up my conclusion here: https://github.com/graphprotocol/graph-node/issues/571#issuecomment-446038693

PR #185, opened 2018-12-12
Don't check mandatory fields for null

leoyvens:
Helps with #169 and https://github.com/graphprotocol/graph-node/issues/571. This solves those issues for the `Boolean!` graphql type, but not for the optional `Boolean`.

Overall this is a good polish over what we have now, so worth doing regardless of what the final solution to those issues is.


leodasvacas:
This has been updated to implement the solution we considered to be least bad: Which is to allow `false` and `0` to be returned if the field is missing, requiring the user to use `isSet` to disambiguate. See https://github.com/graphprotocol/graph-ts/pull/31.

Jannis (changes requested):
Great! One remark: The generated types for the example subgraphs are in `src/`. Can you rerun codegen with `--output-dir=src/types`?

leodasvacas:
@Jannis Ah well noticed, I did find it strange that the generated types weren't there already.

Jannis (approved):

PR #187, opened 2018-12-21
Don't default to port 8020 if no port is set

Jannis:
The problem is: port 80 will result in no port being set and we don't
want to default to 8020 in that case. So it's probably best to just
assume no port means port 80 and use that.


nenadjaja (approved):
👍

PR #202, opened 2019-01-19
Allow graph-node to provide full URLs

timmclean:
See also: https://github.com/graphprotocol/graph-node/pull/699

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -351,12 +351,26 @@ app
           if (!requestError && !jsonRpcError) {
             logger.status('Deployed successfully')
 
-            // Assume that the host is the same.
-            // In the future the deployment router should also return the host.
-            let base = requestUrl.protocol + '//' + requestUrl.hostname
-            logger.status('Playground:        ', base + res.playground)
-            logger.status('Queries (HTTP):    ', base + res.queries)
-            logger.status('Subscriptions (WS):', base + res.subscriptions)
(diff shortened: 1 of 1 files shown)


graph.js
```diff
@@ -351,12 +351,26 @@ app
...
-            logger.status('Queries (HTTP):    ', base + res.queries)
-            logger.status('Subscriptions (WS):', base + res.subscriptions)
+            const base = requestUrl.protocol + '//' + requestUrl.hostname
+
+            let playground    = res.playground
+            let queries       = res.queries
+            let subscriptions = res.subscriptions
+
+            // Add a base URL if graph-node did not return the full URL
+            if (playground.charAt(0) === ':') {
```
Jannis: Not the most robust of checks... Perhaps trying to parse this as an URL and falling back to base + relative if that fails would be better.

Jannis (changes requested):

graph.js
timmclean: This is to be backwards compatible with the existing behaviour. Right now, graph-node always returns routes prefixed with `:`. In https://github.com/graphprotocol/graph-node/pull/699 , routes will be prefixed will start with `:` if they are incomplete and need to have the protocol and host added

Jannis: Ah, yes, I understand this now.

Jannis (approved):

PR #208, opened 2019-02-13
Make it harder to use unprefixed subgraph names, add advice to graph deploy errors

### LINKED ISSUE #205: Fail if `graph init` is called with a subgraph name with no account name prefix
It's confusing people a little bit that we require subgraph names with prefixes on the hosted service but not in `graph init`.

We could add a flag to allow `graph init` to pass even without a name prefix for people who develop locally. But the prefix requirement should be the default.

Jannis:
Resolves #205 and #206.

## Running `graph init` with an unprefixed vs. prefixed subgraph name

image

## Running `graph deploy` for a non-existent subgraph

image

### DIFF
--- a/graph.js
+++ b/graph.js
@@ -329,6 +329,8 @@ app
 
     let logger = new Logger(0, { verbosity: getVerbosity(app) })
 
+    let hostedService = cmd.node.match(/thegraph.com/)
+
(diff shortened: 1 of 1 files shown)


leoyvens (approved):
Nice user experience improvements

PR #224, opened 2019-03-07
Add graph init with --from-example and --from-contract modes

Jannis:
This supersedes #216 by supporting both the old and the new `graph init` via

* `--from-example` and 
* `--from-contract <address>`

mode flags.

## TODO

- [x] Update README
- [x] Update docs on https://thegraph.com/docs/

## Example usage

If all arguments are provided for either mode, `graph init` will run in non-interactive mode:

```sh
# From the example subgraph:
graph init --from-example user/subgraph-name output-dir
graph init --from-example --allow-simple-name subgraph-name output-dir

# From an existing contract:
graph init \
  --from-contract 0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4d \
  --network mainnet \
  user/subgraph-name \
  output-dir
```

If one or more arguments are not provided, `graph init` will run the user through an interactive form to create a subgraph from an existing contract, pre-populating fields with the provided values:

```sh
$ graph init --from-contract 0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4d jannis/decentraland-land
✔ Subgraph name · jannis/decentraland-land
✔ Directory to create the subgraph in · land
✔ Ethereum network · mainnet
✔ Contract address · 0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4d
✔ Fetching ABI from Etherscan
———
  Generate subgraph from ABI
  Write subgraph to directory
✔ Create subgraph scaffold
✔ Initialize subgraph repository
✔ Install dependencies with yarn
✔ Generate ABI and schema types with yarn codegen

Subgraph jannis/decentraland-land created in land

Next steps:

  1. Run `graph auth https://api.thegraph.com/deploy/ <access-token>`
     to authenticate with the hosted service. You can get the access token from
     https://thegraph.com/explorer/dashboard/.

  2. Type `cd land` to enter the subgraph.

  3. Run `yarn deploy` to deploy the subgraph to
     https://thegraph.com/explorer/subgraph/jannis/decentraland-land.

Make sure to visit the documentation on https://thegraph.com/docs/ for further information.
```

If an ABI cannot be found on Etherscan, this form asks for a local ABI file path:

```sh
$ graph init --from-contract 0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4f jannis/decentraland-land
✔ Subgraph name · jannis/decentraland-land
✔ Directory to create the subgraph in · land-2
✔ Ethereum network · mainnet
✔ Contract address · 0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4f
✖ Failed to fetch ABI from Etherscan: Unexpected token C in JSON at position 0
✔ ABI file (path) · /Users/jannis/Work/graphprotocol/subgraphs/decentraland/abis/LANDRegistry.json
———
  Generate subgraph from ABI
  Write subgraph to directory
✔ Create subgraph scaffold
✔ Initialize subgraph repository
⠋ Install dependencies with yarn
...
```

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -274,6 +274,7 @@ module.exports = {
       subgraphName,
     })
 
+    // Exit immediately when the form is cancelled
     if (inputs === undefined) {
       process.exit(1)
       return
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -189,6 +189,13 @@ module.exports = {
       return
     }
 
+    // Fail if the ABI does not contain any events
+    if (abi.filter(item => item.type === 'event').length === 0) {
+      print.error(`ABI does not contain any events`)
+      process.exitCode = 1
+      return
+    }
+
     // Fail if the output directory already exists
     if (toolbox.filesystem.exists(directory)) {
       print.error(`Directory or file "${directory}" already exists`)
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -333,6 +333,8 @@ const initRepository = async (toolbox, directory) =>
     `Initialize subgraph repository`,
     `Failed to initialize subgraph repository`,
     async spinner => {
+      // Remove .git dir in --from-example mode; in --from-contract, we're
+      // starting from an empty directory
       let gitDir = path.join(directory, '.git')
       if (toolbox.filesystem.exists(gitDir)) {
         await toolbox.filesystem.remove(gitDir)
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -104,7 +104,9 @@ const processInitForm = async (toolbox, subgraphName, { allowSimpleName }) => {
 
 const loadAbiFromEtherscan = async (network, address) => {
   let result = await fetch(
-    `https://api.etherscan.io/api?module=contract&action=getabi&address=${address}`
+    `https://${
+      network === 'mainnet' ? 'api' : `api-${network}`
+    }.etherscan.io/api?module=contract&action=getabi&address=${address}`
   )
   let json = await result.json()
   return JSON.parse(json.result)
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -1,10 +1,14 @@
 const chalk = require('chalk')
+const fetch = require('node-fetch')
 const os = require('os')
 const path = require('path')
+const toolbox = require('gluegun/toolbox')
 const {
   getSubgraphBasename,
   validateSubgraphName,
 } = require('../command-helpers/subgraph')
+const { withSpinner, step } = require('../command-helpers/spinner')
+const { generateScaffold, writeScaffold } = require('../scaffold')
 
 const HELP = `
 ${chalk.bold('graph init')} [options] ${chalk.bold('<subgraph-name>')}
@@ -12,8 +16,109 @@ ${chalk.bold('graph init')} [options] ${chalk.bold('<subgraph-name>')}
 ${chalk.dim('Options:')}
 
   -h, --help                    Show usage information
-      --allow-simple-name       Use a subgraph name without a prefix component (default: false)
+      --allow-simple-name       Use a subgraph name without a prefix (default: false)
 `
+
+const processInitForm = async (toolbox, subgraphName, { allowSimpleName }) => {
+  let useEtherscan = true
+  let questions = [
+    {
+      type: 'input',
+      name: 'subgraphName',
+      message: 'Subgraph name',
+      initial: subgraphName,
+      validate: name => {
+        try {
+          validateSubgraphName(name, { allowSimpleName })
+          return true
+        } catch (e) {
+          return `${e.message}
+
+  Examples:
+  
+    $ graph init ${os.userInfo().username}/${name}
+    $ graph init ${name} --allow-simple-name`
+        }
+      },
+      result: value => {
+        subgraphName = value
+        return value
+      },
+    },
+    {
+      type: 'input',
+      name: 'directory',
+      message: 'Directory to create the subgraph in',
+      initial: () => getSubgraphBasename(subgraphName),
+      validate: value =>
+        toolbox.filesystem.exists(value) ? 'Directory already exists' : true,
+    },
+    {
+      type: 'select',
+      name: 'network',
+      message: 'Ethereum network',
+      choices: ['mainnet', 'kovan', 'rinkeby', 'ropsten'],
+      initial: 'mainnet',
+    },
+    {
+      type: 'input',
+      name: 'address',
+      message: 'Contract address',
+      validate: value => {
+        // Validate whether the address is valid
+        let pattern = /^(0x)?[0-9a-fA-F]{40}$/
+        if (pattern.test(value)) {
+          return true
+        } else {
+          return `Contract address "${value}" is invalid.
+  Must be 40 hexadecimal characters, with an optional '0x' prefix.`
+        }
+      },
+    },
+    {
+      type: 'confirm',
+      name: 'useEtherscan',
+      message: 'Fetch ABI from Etherscan?',
+      initial: useEtherscan,
+      result: value => {
+        useEtherscan = value
+        return value
+      },
+    },
+    {
+      type: 'input',
+      name: 'abiFile',
+      message: 'ABI file (path)',
+      skip: () => useEtherscan,
+      validate: value =>
+        toolbox.filesystem.exists(value) ? true : 'File does not exist',
+    },
+  ]
+
+  try {
+    return await toolbox.prompt.ask(questions)
+  } catch (e) {
+    return undefined
+  }
+}
+
+const loadAbiFromEtherscan = async (network, address) => {
+  let result = await fetch(
+    `https://api.etherscan.io/api?module=contract&action=getabi&address=${address}`
+  )
+  let json = await result.json()
+  return JSON.parse(json.result)
+}
+
+const loadAbiFromFile = async file => await toolbox.filesystem.read(abiFile, 'json')
+
+const loadAbi = async ({ abiFile, address, network, useEtherscan }) => {
+  if (useEtherscan) {
+    return await loadAbiFromEtherscan(network, address)
+  } else {
+    return await loadAbiFromFile(abiFile)
+  }
+}
 
 module.exports = {
   description: 'Creates a new subgraph with basic scaffolding',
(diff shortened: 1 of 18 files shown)


src/scaffold.js
```diff
@@ -0,0 +1,193 @@
...
+
+// Mapping
+
+const generateMapping = ({ abi, subgraphName }) =>
+  prettier.format(
+    `
+  import { ${abiEvents(abi).map(
+    event => `${event.name} as ${event.name}Event`
+  )}} from '../generated/Contract/Contract'
+  import { ${abiEvents(abi).map(event => event.name)} } from '../generated/schema'
```
leodasvacas: This assumes that generated code lives in `/generated`, which would be a good name, but currently codegen puts things in `/types`.

leodasvacas:
The "Next steps:" instructions don't instruct me to run codegen, which is necessary before deploy.

Jannis:
@leodasvacas That's because `graph init` runs it for you 😉

leodasvacas:
@Jannis ah nice, I got confused because of the `/generated` issue.

src/scaffold.js
Jannis: Not on master. I switched to `generated` as part of the Gluegun port, although now I feel bad because it's a breaking change. But the generated code here will work.

src/commands/init.js
```diff
@@ -1,64 +1,203 @@
...
-      print.error(e.message)
-      print.info(`
-Examples:
 
-  $ graph init ${os.userInfo().username}/${subgraphName}
-  $ graph init ${subgraphName} --allow-simple-name`)
-      process.exitCode = 1
-      return
-    }
+    let [subgraphName, directory] = fixParameters(toolbox.parameters, {
```
leodasvacas: I don't understand what `fixParameters` does.

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
-      let pkgJsonFilename = filesystem.path(directory, 'package.json')
-      let pkgJson = await filesystem.read(pkgJsonFilename, 'json')
+    // If all parameters are provided from the command-line,
+    // go straight to creating the subgraph from an existing contract
+    if (fromContract && subgraphName && directory && network) {
+      let abi
+      try {
+        abi = await loadAbiFromEtherscan(network, fromContract)
+      } catch (e) {
+        process.exitCode = 1
```
leodasvacas: should we not print `e` here?

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+      allowSimpleName,
+      directory,
+      address: fromContract,
+      fromExample,
+      network,
+      subgraphName,
+    })
+
+    if (inputs === undefined) {
+      process.exit(1)
```
leodasvacas: Does `inputs === undefined` have a particular meaning? Should we print something here?

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+}
+
+const initRepository = async (toolbox, directory) =>
+  await withSpinner(
+    `Initialize subgraph repository`,
+    `Failed to initialize subgraph repository`,
+    async spinner => {
+      let gitDir = path.join(directory, '.git')
+      if (toolbox.filesystem.exists(gitDir)) {
+        await toolbox.filesystem.remove(gitDir)
```
leodasvacas: How could the git directory already exist, doesn't graph <strike>autodave</strike> init always start from an empty dir?

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+    `Failed to clone example subgraph`,
+    async spinner => {
+      await system.run(
+        `git clone http://github.com/graphprotocol/example-subgraph ${directory}`
+      )
+      return true
+    }
+  )
+  if (!cloned) {
+    process.exitCode = 1
```
leodasvacas: I tried running `graph init --from-example` without internet connection and it failed silently, probably because of this. We should print something, and maybe also throw an exception so we can also print a more general error?

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+  let installed = await installDependencies(toolbox, directory, commands.install)
+  if (installed !== true) {
+    process.exitCode = 1
+    return
+  }
+
+  // Run code-generation
+  let codegen = await runCodegen(toolbox, directory, commands.codegen)
+  if (codegen !== true) {
+    process.exitCode = 1
```
leodasvacas: Setting the exit code and then returning doesn't seem like the best practice, perhaps bubble up an exception and then set the exit code when catching it?

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+  // Fail if the output directory already exists
+  if (toolbox.filesystem.exists(directory)) {
+    print.error(`Directory or file "${directory}" already exists`)
+    process.exitCode = 1
+    return
+  }
+
+  if (abiEvents(abi).length === 0) {
+    // Fail if the ABI does not contain any events
+    print.error(`ABI does not contain any events`)
```
leodasvacas: should an ABI with no events fail everything? Since we'll soon get other trigger types, it may make sense to scaffold as much as we can even with no events.

src/commands/init.js
```diff
@@ -83,84 +222,323 @@ Examples:
...
+  let installed = await installDependencies(toolbox, directory, commands.install)
+  if (installed !== true) {
+    process.exitCode = 1
+    return
+  }
+
+  // Run code-generation
+  let codegen = await runCodegen(toolbox, directory, commands.codegen)
+  if (codegen !== true) {
+    process.exitCode = 1
```
leodasvacas: Same comment as above about error handling.

src/scaffold.js
```diff
@@ -0,0 +1,193 @@
...
+
+const ethereumTypeToGraphQL = name => {
+  let ascType = ascTypeForEthereum(name)
+  return valueTypeForAsc(ascType)
+}
+
+const generateEventType = event => `type ${event.name} @entity {
+      id: ID!
+      ${event.inputs
+        // TODO: Support inputs without names
```
leodasvacas: Out of curiosity, what sort of inputs have no name? How does this code behave in that case?

leoyvens (changes requested):
A bunch of small comments. Nice tests! It would be cool to also have a test for the generated contents.

src/commands/init.js
Jannis: Basically, if you run `graph init --from-example leodasvacas/some-subgraph`, you would expect `toolbox.parameters.first` to be `'leodasvacas/some-subgraph'` and `toolbox.parameters.array` to be `['leodasvacas/some-subgraph']`. However, due to the way Gluegun parses command line arguments, it actually considers `leodasvacas/some-subgraph` to be the value for the `--from-example` option, so `fromExample` would be `'leodasvacas/some-subgraph'` and `tools.parameters.first` would be `undefined`. Ouch.

`fixParameters` checks all provided boolean options for whether they have an argument value; if that is the case, it pushes that to the front of the `toolbox.parameters.array` and returns the result. It's a hack to work around the above problem.

I had some fun with this last week: https://github.com/infinitered/gluegun/pull/464

This reminds me though: we should be using `fixParameters` in all our commands, because invocations like `graph deploy --watch leodasvacas/some-subgraph` are also affected. And I should add a comment. 😄

Jannis: Tracked here: https://github.com/graphprotocol/graph-cli/issues/231

Jannis:
Thanks @leodasvacas, I'll address the comments in the morning!

src/commands/init.js
Jannis: With `--from-contract` yes, but not with `--from-example` (where it clones the example subgraph). This code is used in both cases.

src/commands/init.js
Jannis: The `inputs` are `undefined` if the form is cancelled. I've added a comment.

src/commands/init.js
Jannis: The thing is: `runCodegen` already runs a spinner that will log an error if code generation fails. I didn't want to output two identical errors. We could throw an error here and then not print it at the caller level but that is not better IMHO.

src/commands/init.js
Jannis: There was a bug in the form processing that would cause it to silently fail when it shouldn't. This should be fixed now.

src/scaffold.js
Jannis: Oops, that comment is outdated — we fall back to `param${index}` for inputs with no names now.

I don't recall where we encountered event inputs without names, but I'm certain we did somewhere. That's why the code generation does the same — if an input has no name, it uses `param${index}`.

src/commands/init.js
Jannis: It's not obvious, but since `loadAbiFromEtherscan` uses a spinner, the spinner takes care of printing an error when it fails.

src/commands/init.js
Jannis: Not right now though — if we don't fail here, the subgraph would later fail to build because of there being no event handlers defined. We can improve this when the other trigger types are added.

src/commands/init.js
Jannis: Same answer 😁

Jannis:
@leodasvacas I've fixed a few things, added a few comments and tried to address with comments with replies. Let me know what you think.

src/scaffold.js
leodasvacas: So the comment can be removed?

src/commands/init.js
leodasvacas: Ah ok, so maybe comment that this is removing the .git in the `--from-example` case.

src/scaffold.js
Jannis: Yes, I thought I had pushed a commit already. Done that now.

src/commands/init.js
leodasvacas: Where's the comment?

src/commands/init.js
leodasvacas: ah, awesome that you're trying to fix it upstream.

leodasvacas: Did you get to adding a comment?

src/commands/init.js
Jannis: https://github.com/graphprotocol/graph-cli/blob/jannis/graph-init-with-two-modes/src/commands/init.js#L281 😉

src/commands/init.js
Jannis: Done

src/commands/init.js
Jannis: I actually rewrote the whole function and added a comment:

https://github.com/graphprotocol/graph-cli/blob/1bf7373ad7925c48b16288937fdf6f765ab28061/src/command-helpers/gluegun.js#L1-L47

leodasvacas: Ah yes sorry. Great comment!

src/scaffold.js
```diff
@@ -0,0 +1,192 @@
...
+        'create-local': `graph create --node http://localhost:8020/ ${subgraphName}`,
+        'remove-local': `graph remove --node http://localhost:8020/ ${subgraphName}`,
+        'deploy-local':
+          `graph deploy ` +
+          `--node http://localhost:8020/ ` +
+          `--ipfs http://localhost:5001 ` +
+          subgraphName,
+      },
+      dependencies: {
+        '@graphprotocol/graph-cli': '0.7.2',
```
leodasvacas: We'll need to be careful to remember to update this, at least on breaking changes.

leoyvens (dismissed):

src/scaffold.js
Jannis: Indeed. I have an idea: How about we take the package version and put it in here? Then we're always up to date.

leodasvacas: If that's possible that would be great.

leoyvens (approved):

PR #227, opened 2019-03-11
Improvements

Jannis:
@fordN I decided to submit a PR against your PR instead of a review. This addresses all the comments I would've had:

* Shorter / more simple warning messages.
* Use the built-in `spinner.warn` instead of a custom `stopAndPersist` call.
* Improve `warning` object check in `withSpinner`.
* Add empty line after warnings to balance things visually.

Let me know what you think!

Screenshot:

image

### DIFF
--- a/src/command-helpers/spinner.js
+++ b/src/command-helpers/spinner.js
@@ -14,19 +14,27 @@ const step = (spinner, subject, text) => {
   return spinner
 }
 
-// A wrapper for functions (f) that provides a progress spinner and formatted alert messages.
-// For warning message support the f must return an immutable Map with result and warning keys.
+// Executes the function `f` in a command-line spinner, using the
+// provided captions for in-progress, error and failed messages.
+//
+// If `f` throws an error, the spinner stops with the failure message
+//   and rethrows the error.
+// If `f` returns an object with a `warning` and a `result` key, the
+//   spinner stops with the warning message and returns the `result` value.
+// Otherwise the spinner prints the in-progress message with a check mark
+//   and simply returns the value returned by `f`.
 const withSpinner = async (text, errorText, warningText, f) => {
   let spinner = toolbox.print.spin(text)
   try {
     let result = await f(spinner)
-    if(warningText && immutable.Map.isMap(result)) {
-      if (result.get('warning', false)) {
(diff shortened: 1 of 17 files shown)


Jannis:
Hmm, ok, I've decided to bring the `from` back in the warning caption:

image

Jannis:
I'll fix the tests tomorrow.

Jannis:
Tests are fixed — but Node 11 seems to fail on Travis (even on master). Let's ignore that for the moment and look into it when we have time.

src/command-helpers/spinner.js
```diff
@@ -14,19 +14,27 @@ const step = (spinner, subject, text) => {
...
   try {
     let result = await f(spinner)
-    if(warningText && immutable.Map.isMap(result)) {
-      if (result.get('warning', false)) {
-        spinner.stopAndPersist({symbol: '⚠️', text: `${warningText}: ${result.get('warning')}`})
-        return result.get('result')
+    if (typeof result === 'object') {
+      if (result.warning && Object.keys(result).indexOf('result') >= 0) {
+        spinner.warn(`${warningText}: ${result.warning}`)
+        spinner.succeed(text)
```
fordN: Is it necessary to have both `spinner.warn` and `spinner.succeed` when there is a warning?

fordN (changes requested):

src/command-helpers/spinner.js
Jannis: Yep, because we want to see `[x] Load subgraph` even if there were warnings.

Jannis:
@fordN I feel I've addressed your point. Anything else? 😉

src/command-helpers/spinner.js
fordN: My understanding is that the loading text (in this case "Load subgraph from...") is set by the first line, `let spinner = toolbox.print.spin(text)`.   I don't see any difference in the output with the line in question removed.

Jannis: First with the extra `succeed`, then without it:
image

Only in the first example is `✔ Load subgraph from subgraph.yaml` logged after the warnings. And that's what we want.

fordN:
👍

fordN (approved):

PR #229, opened 2019-03-12
Add support for blockHandlers and callHandlers

### LINKED ISSUE #223: Update subgraph validation to block and transaction triggers
* Validate these fields
* Decide on the data to pass down to handlers

Jannis:
Resolves #223.

This adds `blockHandlers` and `callHandlers` the subgraph manifest schema and includes them in the subgraph validation.

Open question: Should we introduce a new mapping kind alias called `ethereum/contract` for `ethereum/event`, since we're now supporting more than events?


Jannis:
Tests are only failing because of Node 11 in Travis acting up. The Node 10 tests pass fine.

manifest-schema.graphql
```diff
@@ -27,14 +27,25 @@ type EthereumContractMapping {
...
 type EthereumContractAbi {
   name: String!
   file: File!
 }
 
+type EthereumBlockHandler {
+  handler: String!
+}
+
+type EthereumTransactionHandlers {
```
olibearo: You mentioned in Slack this should be singular, not plural. Should we also make the change for `EthereumContractEventHandlers`?

manifest-schema.graphql
```diff
@@ -27,14 +27,25 @@ type EthereumContractMapping {
   file: File!
   entities: [String!]!
   abis: [EthereumContractAbi!]!
-  eventHandlers: [EthereumContractEventHandlers]!
+  blockHandlers: EthereumBlockHandler
```
olibearo: Also, maybe this should be singular as well since it is not an array?

Jannis: Yep, good point

manifest-schema.graphql
Jannis: Yeah, that makes sense!

Jannis:
@olibearo This is ready. Do you want to take another look?

ghost (approved):

PR #235, opened 2019-03-26
Codegen for Tuple types

fordN:
A part of the Tuple/Struct support epic: 

This PR adds support for Tuple types in `graph codegen`:
- Update event types code generation to create a class for each Tuple type and to use them in the Tuple param method generation.
- In smart contract class, add support for generating call methods that use Tuples as input or output.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -24,6 +24,7 @@ module.exports = class AbiCodeGenerator {
           'Bytes',
           'Address',
           'BigInt',
+          'Tuple'
         ],
         '@graphprotocol/graph-ts'
       ),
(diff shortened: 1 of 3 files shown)


src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
             name = `param${index}`
           }
           paramsClass.addMethod(
-            tsCodegen.method(
+            input.get('type') === 'tuple' ?
+              tsCodegen.method(
+                `get ${name}`,
+                [],
+                tsCodegen.namedType(name).capitalize(),
```
Jannis: I'd move the `tsCodegen.namedType(name).capitalize()` into a `tupleClassName` variable, so you can reuse it in the generated `return` statement below.

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+                `
+            return <${tsCodegen.namedType(name).capitalize()}>${typesCodegen.ethereumValueToAsc(
```
Jannis: * You could then reuse the `tupleClassName` here.
* The indentation of the `return` here doesn't align with the opening backtick; the closing backtick is off too.
* Can we return `return ... as ${tupleClassName}` instead of using the `<${tupleClassName}>` syntax? I feel it's less cryptic.

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+                  `this._event.parameters[${index}].value`,
+                  input.get('type')
+                )}
+            `
+              ) :
+              tsCodegen.method(
+                  `get ${name}`,
```
Jannis: The indentation here suggests to me that `prettier` didn't run. 😉

src/codegen/abi.js
Jannis: Another thought: we can't name tuples after their parameter names, that'll cause too many naming clashes.

Let's use event name (captialized) + param name or index as the name for the tuple classes.

Jannis: Maybe even append "Struct" to it, so e.g. `TransferParam0Struct` or `TransferParamFooStruct`.

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+                return ${typesCodegen.ethereumValueToAsc(
+                  `this._event.parameters[${index}].value`,
+                  input.get('type')
+                )}
+                `
+              )
+          )
+        })
+
+        // Generate param getters differently for Tuple params
```
Jannis: I'm not sure what the "differently" refers to here. I'd just make this
```
// Generate a class for each tuple parameter
```

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+        // Also generate a class for each Tuple param
+        let tupleKlasses = [];
```
Jannis: We can use "Classes" and "Class" in this part of the code since there's a prefix in the name, so `tupleClasses`, `tupleClass` like with `paramClass`.

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+        event.get('inputs').filter(input => input.get('type') === 'tuple').forEach((input, index) => {
+          let tupleKlass = tsCodegen.klass(tsCodegen.namedType(input.get('name')).capitalize(), {
```
Jannis: Prepend event name and input name/index to the class name here.

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+            export: true,
+            extends: 'Tuple',
+          });
+          input.get('components').forEach((component, index) => {
+            let name = component.get('name')
+            if (name === undefined || name === null || name === '') {
+              name = `param${index}`
```
Jannis: param -> value

src/codegen/abi.js
```diff
@@ -59,18 +60,60 @@ module.exports = class AbiCodeGenerator {
...
+            }
+            tupleKlass.addMethod(
+              tsCodegen.method(
               `get ${name}`,
               [],
-              typesCodegen.ascTypeForEthereum(input.get('type')),
+              typesCodegen.ascTypeForEthereum(component.get('type')),
```
Jannis: Q: Can tuples be nested?

src/codegen/abi.js
```diff
@@ -86,7 +129,7 @@ module.exports = class AbiCodeGenerator {
             `return new ${paramsClassName}(this)`
           )
         )
-        return [klass, paramsClass]
+        return [klass].concat(paramsClass,tupleKlasses)
```
Jannis: Let's do `[klass, paramsClass, ...tupleClasses]` here. Modern JS and stuff 😁

fordN:
I'm closing this because it is a duplicate of #243  which has the latest tuple support work.

PR #236, opened 2019-03-27
Generate schema code for `BigDecimal`

leodasvacas:
Needs to be tested with complete graph-ts support.


Jannis:
@leodasvacas Hah, I actually meant tests here: https://github.com/graphprotocol/graph-cli/blob/master/src/codegen/types/index.test.js

leodasvacas:
@Jannis added the codegen tests and fixed the dir name.

leodasvacas:
@Jannis I can't merge because of CI, which is failing because the example subgraph's graph-ts doesn't have BigDecimal.

Jannis:
Ah, that makes sense. Let's merge this now and update the example subgraph later when we're ready to release a new graph-ts (waiting for the struct/tuple support and `ipfs.map` being merged).

Jannis:
I'll force merge it now. :godmode:

PR #237, opened 2019-03-28
Add migrations to Graph CLI, bump mappings from 0.0.1 to 0.0.2 automatically

### LINKED ISSUE #233: Auto-bump mapping API version
If a subgraph has defined a mapping API version that is behind the version of `graph-ts` being used: auto-bump the mapping API version in the subgraph manifest to the compatible version. 
  
Perform the bump of the mapping API version during `build`.

Jannis:
Resolves #233.

This is a prerequisite to us releasing new versions of graph-ts and graph-cli.

When running `graph codegen`, `graph build` or `graph codegen`, migrations are now applied to subgraphs. The first migration that is added in this PR automatically bumps `apiVersion` of mappings from `0.0.1` to `0.0.2`.

In action: First, running `graph codegen` before the migration has run (note the `Apply migrations` messages):
```sh
$ graph codegen
  Apply migration: Bump mapping apiVersion from 0.0.1 to 0.0.2
✔ Apply migrations
⚠ Warnings while loading subgraph from subgraph.yaml: Warnings in subgraph.yaml:

    Path: description
    The description is still the one from the example subgraph.
    Please update it to tell users more about your subgraph.

✔ Load subgraph from subgraph.yaml
  Load contract ABI from abis/Gravity.json
✔ Load contract ABIs
  Generate types for contract ABI: Gravity (abis/Gravity.json)
  Write types to generated/Gravity/Gravity.ts
✔ Generate types for contract ABIs
✔ Load GraphQL schema from schema.graphql
  Write types to generated/schema.ts
✔ Generate types for GraphQL schema

Types generated successfully
```

Second, running `graph codegen` after the migration has run (note the `Skip migration` message):
```
$ graph codegen
  Skip migration: Bump mapping apiVersion from 0.0.1 to 0.0.2
✔ Apply migrations
⚠ Warnings while loading subgraph from subgraph.yaml: Warnings in subgraph.yaml:

    Path: description
    The description is still the one from the example subgraph.
    Please update it to tell users more about your subgraph.

✔ Load subgraph from subgraph.yaml
  Load contract ABI from abis/Gravity.json
✔ Load contract ABIs
  Generate types for contract ABI: Gravity (abis/Gravity.json)
  Write types to generated/Gravity/Gravity.ts
✔ Generate types for contract ABIs
✔ Load GraphQL schema from schema.graphql
  Write types to generated/schema.ts
✔ Generate types for GraphQL schema

Types generated successfully
```

And the `git diff` output after running the above:

```diff
git diff subgraph.yaml
diff --git a/subgraph.yaml b/subgraph.yaml
index dbebe88..b115478 100644
--- a/subgraph.yaml
+++ b/subgraph.yaml
@@ -12,7 +12,7 @@ dataSources:
       abi: Gravity
     mapping:
       kind: ethereum/events
-      apiVersion: 0.0.1
+      apiVersion: 0.0.2
       language: wasm/assemblyscript
       entities:
         - Gravatar
```

### DIFF
--- a/src/migrations/bump-mapping-api-version.js
+++ b/src/migrations/bump-mapping-api-version.js
@@ -1,13 +1,41 @@
 const fs = require('fs-extra')
+const path = require('path')
+const semver = require('semver')
 const toolbox = require('gluegun/toolbox')
 const yaml = require('js-yaml')
+
+const getGraphTsVersion = sourceDir => {
+  let pkgJsonFile = path.join(
+    sourceDir,
+    'node_modules',
+    '@graphprotocol',
+    'graph-ts',
+    'package.json',
+  )
+  let data = fs.readFileSync(pkgJsonFile)
+  let jsonData = JSON.parse(data)
+  return jsonData.version
+}
 
 // If any of the manifest apiVersions are 0.0.1, replace them with 0.0.2
 module.exports = {
   name: 'Bump mapping apiVersion from 0.0.1 to 0.0.2',
-  predicate: async ({ manifestFile }) => {
+  predicate: async ({ sourceDir, manifestFile }) => {
+    // Obtain the graph-ts version, if possible
+    let graphTsVersion
+    try {
+      graphTsVersion = getGraphTsVersion(sourceDir)
+    } catch (_) {
+      // If we cannot obtain the version, return a hint that the graph-ts
+      // hasn't been installed yet
+      return 'graph-ts dependency not installed yet'
+    }
+
     let manifest = yaml.safeLoad(fs.readFileSync(manifestFile, 'utf-8'))
     return (
+      // Only migrate if the graph-ts version is > 0.5.1...
+      semver.gt(graphTsVersion, '0.5.0') &&
+      // ...and we have a manifest with mapping > apiVersion = 0.0.1
       manifest &&
       typeof manifest === 'object' &&
       Array.isArray(manifest.dataSources) &&
--- a/src/migrations/bump-mapping-api-version.js
+++ b/src/migrations/bump-mapping-api-version.js
@@ -1,0 +1,42 @@
+const fs = require('fs-extra')
+const toolbox = require('gluegun/toolbox')
+const yaml = require('js-yaml')
+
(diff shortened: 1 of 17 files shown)


leodasvacas:
Looks good. An input -> output test might be nice but I don't know how feasible that is. Ideally, we'd run this migration based on the graph-ts version, rather than running it based on the graph-cli version as this is doing right now. But they're usually bumped together so in practice this also works.

Jannis:
@leodasvacas Yeah, I've got full I/O tests in the back of my mind but it would be a lot of work right now.

We could check the `@graphprotocol/graph-ts` version in `package.json` as well if that makes sense to you. Is that what you're suggesting: `graph-ts >= x.y.z` + `apiVersion: 0.0.1` -> run the migration?

leodasvacas:
@Jannis exactly, that would be good. We can apply the migration regardless of the current `apiVersion`, to cover the case where user is going from a newer version to an older one.

Jannis:
@leodasvacas I don't like applying the migration regardless of the current `apiVersion`, because the patching isn't granular enough — it operates on the entire file and if we ever add `apiVersion: 0.0.1` anywhere else, it would overwrite that.

I've made the migration stricter now: Only migrate mappings from 0.0.1 to 0.0.2 if the graph-ts dependency is installed and > `0.5.1` (our current version). So the next time we release a new graph-ts version, the migration would kick in, but not with 0.5.1 yet.

src/migrations/bump-mapping-api-version.js
```diff
@@ -1,13 +1,41 @@
...
+    } catch (_) {
+      // If we cannot obtain the version, return a hint that the graph-ts
+      // hasn't been installed yet
+      return 'graph-ts dependency not installed yet'
+    }
+
     let manifest = yaml.safeLoad(fs.readFileSync(manifestFile, 'utf-8'))
     return (
+      // Only migrate if the graph-ts version is > 0.5.1...
+      semver.gt(graphTsVersion, '0.5.0') &&
```
leodasvacas: This isn't true for `0.5.1`, right?

leoyvens (approved):

PR #241, opened 2019-04-02
Map types of indexed array/string parameters to bytes32

Jannis:
This is the Graph CLI part of solving https://github.com/graphprotocol/graph-node/issues/731 / https://github.com/paritytech/ethabi/issues/146.

To deal with indexed parameters properly, we need to treat them as `bytes32` and generate code accordingly.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -25,7 +25,7 @@ module.exports = class AbiCodeGenerator {
           'Address',
           'BigInt',
         ],
-        '@graphprotocol/graph-ts'
+        '@graphprotocol/graph-ts',
       ),
     ]
   }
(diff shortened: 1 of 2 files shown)


src/codegen/abi.js
```diff
@@ -242,27 +247,40 @@ module.exports = class AbiCodeGenerator {
...
 
     return [...types, klass]
   }
+
+  _indexedInputType(inputType) {
+    // strings, bytes and arrays are encoded and hashed to a bytes32 value
+    if (
+      inputType === 'string' ||
+      inputType === 'bytes' ||
+      inputType.match(/\[[0-9]*\]$/g)
```
leodasvacas: I assume that this is matching arrays, but from the regex it's not obvious to me.

leoyvens (approved):

src/codegen/abi.js
Jannis: You assume correctly! I'll add a comment before I merge.

leoyvens (approved):

PR #243, opened 2019-04-04
Codegen for access helpers to Tuple data structures

fordN:
Part of the Solidity structs epic: [graph-node issue #766](https://github.com/graphprotocol/graph-node/issues/766)

This PR updates `graph codegen` to support `Tuple` access in the mappings.  It now generates:
  - A class per Tuple with accessors for each of it's parameters. 
  - Tuple type accessors on the event that convert to a Tuple and reference one of the generated Tuple classes.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -19,12 +19,12 @@ module.exports = class AbiCodeGenerator {
           'JSONValue',
           'TypedMap',
           'Entity',
+          'EthereumTuple',
 
           // AssemblyScript types
           'Bytes',
           'Address',
           'BigInt',
-          'Tuple'
         ],
         '@graphprotocol/graph-ts',
       ),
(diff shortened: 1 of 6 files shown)


Jannis:
How is this different from #235?

src/codegen/abi.js
```diff
@@ -54,28 +56,15 @@ module.exports = class AbiCodeGenerator {
...
-            ),
+          // Generate getters and classes for event params
+          let paramObject = this._generateEventParam(
+            input,
+            index,
+            eventClassName,
+            `event`,
           )
+          paramsClass.addMethod(paramObject.getter)
+          tupleClasses.push(...paramObject.classes)
```
Jannis: Hehe, I guess this could also just be `tupleClasses.concat(paramObject.classes)`.

fordN: This allows us to change it in place though and not create a new reference :).

Jannis: Ah, good point!

Issue #168, opened 2018-11-11
Building subgraphs fails when there are overloaded events and functions


davekaj:
Solidity allows for multiple functions with the same name, but different arguments (overloaded functions). See https://github.com/aragon/aragonOS/blob/dev/contracts/acl/ACL.sol for a good example , they use `hasPermission` three times. 

This causes our compiler to fail because we have duplicate identifiers. 

Short term fix is rename the types created. 

What is the best way to fix this in the future? Use the function signature somehow?

davekaj:
This might be causing errors for Aragon. They tried to build the subgraph, but got errors, but I am not sure what the error is. This would cause it to fail though for sure. We should fix this so they can use it

Jannis:
This also needs to account for events that have the same name as the types we generate for other events, see https://github.com/graphprotocol/graph-cli/issues/167.

PR #247, opened 2019-04-24
De-duplicate event names and contract function names

leodasvacas:
The fix in this PR is sufficient to fix the name duplication issues for event classes and contract methods observed in [the decentraland subgraph](https://github.com/graphprotocol/decentraland-subgraph), by using a numbering scheme for events and methods with the same name.

To fully address #168 there is still the issue of having `event SetPermission` and `event SetPermissionParams`, which is a total edge case. There I'm reluctant of going for a numbering scheme, which I think would over complicate the situation. Since `...Params` types are not usually imported into mappings, it would be mostly backwards compatible to rename them. So can I fix this by just changing the suffix to `...EventParams`?

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -36,10 +36,20 @@ module.exports = class AbiCodeGenerator {
   }
 
   _generateEventTypes() {
+    // Keep a map of how many events were generated so far with the given name.
+    // If there are no name collisions, all entries map to 1.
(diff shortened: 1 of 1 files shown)


Jannis:
@leodasvacas Yes, `...SetPermissionEventParams` sounds fine to me.

Jannis:
Could even do `SetPermission__Params` like we do with return types of function calls.

PR #251, opened 2019-04-25
make keytar an optional dependency

iameli:
it's sort of a shame to have all projects that make use of graph-cli
require libsecret when it's not necessary in the vast majority of cases.
this commit makes it an optional dependency and doesn't attempt to load
it until it's needed.

i don't anticipate this being a major problem - most of the time this
auth store is being used, it's in a desktop environment where libsecret
will already be present. conversely, most of the time libsecret is
absent it will be in a CI environment of some sort, where it's likely
not needed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,11 +15,13 @@
     "ipfs-http-client": "^28.1.2",
     "jayson": "^2.0.6",
     "js-yaml": "^3.12.0",
-    "keytar": "^4.3.0",
     "node-fetch": "^2.3.0",
     "pkginfo": "^0.4.1",
     "prettier": "^1.13.5",
     "request": "^2.88.0"
+  },
+  "optionalDependencies": {
+    "keytar": "^4.3.0"
   },
   "bin": {
     "graph": "bin/graph"
--- a/src/command-helpers/auth.js
+++ b/src/command-helpers/auth.js
@@ -1,6 +1,19 @@
-const keytar = require('keytar')
 const toolbox = require('gluegun/toolbox')
 const { normalizeNodeUrl } = require('./node')
+
+// keytar is listed as an optional dependency in package.json and
+// dynamically required in this function
+const requireKeytar = () => {
+  try {
+    return require('keytar')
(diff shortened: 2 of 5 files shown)


package.json
```diff
@@ -15,12 +15,14 @@
     "ipfs-http-client": "^28.1.2",
     "jayson": "^2.0.6",
     "js-yaml": "^3.12.0",
-    "keytar": "^4.3.0",
     "node-fetch": "^2.3.0",
     "pkginfo": "^0.4.1",
     "prettier": "^1.13.5",
     "request": "^2.88.0"
   },
+  "optionalDependencies": {
```
Jannis: Guess what: I didn't even know optional dependencies were a thing in NPM/Yarn. 😳

src/command-helpers/auth.js
```diff
@@ -1,7 +1,20 @@
...
 const { normalizeNodeUrl } = require('./node')
 
+// keytar is listed as an optional dependency in package.json and
+// dynamically required in this function
+const requireKeytar = () => {
+  try {
+    return require('keytar')
+  } catch (e) {
+    throw new Error(
+      `require(keytar) failed: ${e.message}. on linux, this is often` +
```
Jannis: I'd like to change this message a little bit: make it less "technical", mention the dev packages for libsecret (`libsecret-1-dev` / `libsecret-devel`) because the regular package isn't enough, and improve the capitaliztion a little. How about this?

```
Cannot store the access token because dependencies are missing. If you are
on Linux, try installing `libsecret-1-dev` (Debian, Ubuntu etc.) or
`libsecret-devel` (RedHat, Fedora etc.) and reinstalling Graph CLI afterwards.
```

Jannis:
@iameli Nevermind, I've decided to merge this as it is and update the error message afterwards. We're reparing new releases right now and I'd like to include this.

package.json
iameli: Yeah, they're kind of odd but pretty good for this sort of use case.

PR #252, opened 2019-04-25
General method for deduping methods

leodasvacas:
This fixes the (future) Uniswap issues and is also a more elegant fix for the decentraland issue.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -200,7 +200,6 @@ module.exports = class AbiCodeGenerator {
       ),
     )
 
-    let generatedMethods = new Map()
     this.abi.data.forEach(member => {
       switch (member.get('type')) {
         case 'function':
@@ -209,13 +208,6 @@ module.exports = class AbiCodeGenerator {
             member.get('stateMutability') === 'pure'
           ) {
             let methodName = member.get('name')
-            let currentCount = generatedMethods.get(methodName)
-            if (currentCount === undefined) {
-              generatedMethods.set(methodName, 1)
-            } else {
-              generatedMethods.set(methodName, currentCount + 1)
-              methodName += currentCount
-            }
             
             // Generate a type for the result of calling the function
             let returnType = undefined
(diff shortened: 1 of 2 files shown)


leodasvacas:
> We would generate doSomething and doSomething1 for the overloaded variants, and then doSomething11 for the original doSomething1 method. That is awkward.

Actually in that case this PR would not generate `doSomething11` and instead generate `doSomething1` and cause a conflict.

It's a good suggestion to look at all the names at once, but I didn't follow exactly what names you'd prefer having generated. Perhaps a nice alternative would be to generate `doSomething`, `doSomething01` for the overload and keep `doSomething1` intact.

Before diving deeper, I'd like to propose merging this as-is, since it solves problems we're having right now. There are edge cases in which we still fail, but it's not obvious how to best solve them and maybe no contracts that hit that edge case even exist, so we can open an issue for that and revisit if ever necessary.

Jannis:
@leodasvacas Maybe it wasn't so easy to follow. 😄 

I was basically advocating for what you are saying too, to leave the original method names intact (e.g. if there is `doSomething1` in the ABI, keep it as `doSomething1` and generate enumerated names for overloaded methods around the existing names.

I'm not against merging it now, just a little hesitant to introduce behavior that we later break again by generating different names (if suddenly `doSomething1` meant something different, that would break existing subgraphs; and if the signature somehow ends up being the same, you wouldn't even notice that suddenly you're calling a different function). 

If what you're saying is true though and the currently solution wouldn't even allow a subgraph with `doSomething(address)`, `doSomething(uint256)` and `doSomething1(uint256)` to compile, then I think we're on the safe side.

---

I just checked the code and indeed, it would not touch the existing `doSomething1`, as there would be no other instances of `doSomething1`, so no enumeration would happen. That's ok for now.

tl;dr Let's merge 😁

leodasvacas:
@Jannis Your point about breaking changes is a good one. But we're in agreement.

PR #255, opened 2019-05-05
Fix how graph init handles --abi

Jannis:
This bug was reported earlier on Discord. Previously, when `--abi` was provided, neither was the value filled into the form correctly, nor was it loaded if all parameters were provided. This fixes both.

This also adds a test to verify that `graph init` completes successfully when provided with the `--abi` flag.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -122,22 +122,18 @@ const processInitForm = async (
       type: 'input',
       name: 'abi',
       message: 'ABI file (path)',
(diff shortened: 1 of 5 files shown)


leodasvacas:
To confirm, this fixes #256, yes?

leoyvens (approved):

Jannis:
@leodasvacas Yes, indeed! I didn't even see that issue.

Issue #257, opened 2019-05-05
EthereumTuple ERROR


tobyjaguar:
I was able to build the subgraph directory for an existing contract but receive the following error when trying to deploy:

`Compile subgraphERROR TS2305: Module '~lib/@graphprotocol/graph-ts' has no exported member 'EthereumTuple'.`

The build error logs:
```
EthereumTuple,
   ~~~~~~~~~~~~~
 in generated/Contract/Contract.ts(8,2)

ERROR TS2377: Constructors for derived classes must contain a 'super' call.

 export class Entity extends TypedMap<string, Value> {
  constructor() {
    this.entries = new Array<TypedMapEntry<string, Value>>(0)
  }
                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 in ~lib/@graphprotocol/graph-ts/index.ts(810,53)

✖ Failed to compile subgraph: Failed to compile data source mapping: Compile error
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

This contract uses the experimental `ABIEncoderV2` in which we return structs.
Is this feature supported with thegraph?

PaulRBerg:
Structs don't seem to be supported: https://github.com/graphprotocol/graph-node/issues/766

PaulRBerg:
In my case, I got this error output:

```
ERROR TS2305: Module '~lib/@graphprotocol/graph-ts' has no exported member 'EthereumTuple'.

   EthereumTuple,
   ~~~~~~~~~~~~~
in src/types/Sablier/Sablier.ts(8,2)

ERROR TS2377: Constructors for derived classes must contain a 'super' call.

export class Entity extends TypedMap<string, Value> {
  constructor() {
    this.entries = new Array<TypedMapEntry<string, Value>>(0)
}
                                                      ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 in ~lib/@graphprotocol/graph-ts/index.ts(810,53)
```

I'm not using the ABIEncoderV2 so I don't think this should fail. I am, though, instantiating structs in my [contract][streams-contract]'s constructor.

My environment:

- macOS
[...]

PR #260, opened 2019-05-08
graph init: Disambiguate names, spread tuples over multiple entity fields

### LINKED ISSUE #253: Conversion from 'AssemblyScript' to 'Value' for source type 'EthereumTuple' is not supported
When trying load in an ABI local file for a deployed contract, I get the following error:

```
Error: Conversion from 'AssemblyScript' to 'Value' for source type 'EthereumTuple' is not supported
    at findConversionFromType (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/codegen/types/index.js:42:11)
    at valueTypeForAsc (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/codegen/types/index.js:96:3)
    at ethereumTypeToGraphQL (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/scaffold.js:89:10)
    at event.inputs.map (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/scaffold.js:97:50)
    at Array.map (<anonymous>)
    at generateEventType (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/scaffold.js:95:10)
    at Array.map (<anonymous>)
    at generateSchema (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/scaffold.js:107:8)
    at generateScaffold (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/scaffold.js:152:16)
    at withSpinner (/Users/brockelmore/.nvm/versions/node/v8.12.0/lib/node_modules/@graphprotocol/graph-cli/src/commands/init.js:529:28)
```

Jannis:
Resolves #253.

Apart from adding a reusable name disambiguation helper for events, functions, event inputs and function inputs and outputs, this adds the following:

1. Entity type names, event names and event handlers generated by `graph init` are now disambiguated using the same logic as is used by `graph codegen`.

2. Event parameters that are of type `tuple` are now translated to multiple fields on the generated event entity type.

### DIFF
--- a/src/scaffold.js
+++ b/src/scaffold.js
@@ -6,8 +6,23 @@ const pkginfo = require('pkginfo')(module)
 const { getSubgraphBasename } = require('./command-helpers/subgraph')
 const { step } = require('./command-helpers/spinner')
 const { ascTypeForEthereum, valueTypeForAsc } = require('./codegen/types')
+const ABI = require('./abi')
+const util = require('./codegen/util')
 
-const abiEvents = abi => abi.filter(item => item.type === 'event')
+const abiEvents = abi => {
+  abi = ABI.normalized(abi)
+  if (abi === null || abi === undefined) {
+    throw new Error('Invalid ABI')
+  }
+  return util.disambiguateNames({
+    values: abi.filter(item => item.type === 'event'),
+    getName: event => event.name,
+    setName: (event, name) => {
+      event._alias = name
+      return event
+    },
+  })
+}
 
 // package.json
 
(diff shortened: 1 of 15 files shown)


src/scaffold.js
```diff
@@ -60,11 +76,11 @@ dataSources:
       abi: Contract
     mapping:
       kind: ethereum/events
-      apiVersion: 0.0.1
+      apiVersion: 0.0.2
```
leodasvacas: Since we're doing this big PR to fix `graph init` issue before the release, should we also update the `'@graphprotocol/graph-ts'` version in this file?

leodasvacas:
Since we're doing this big PR to fix `graph init` issues before the release, should we also update the `'@graphprotocol/graph-ts'` version that's used in scaffolding? I think that would be the fix for #257.

leoyvens (approved):
Awesome, tests look great. My comment is about something we should also do as part of the release, but it's probably best to merge this and do it as a follow up.

PR #267, opened 2019-05-10
graph init: Improve error when ABIs are not found on Etherscan

### LINKED ISSUE #266: Improve error message when ABIs are not found on Etherscan
When an ABI is not present on Etherscan, Etherscan will return invalid JSON. Right now we just print the JSON parsing error, but really we should print that the ABI is not present.

```sh
➜  subgraphs graph init --from-contract 0x779BEEFFeCD1d26003d1aCE227E2cEe6B41eC54F --network rinkeby jannis/smart-piggies
✔ Subgraph name · jannis/smart-piggies
✔ Directory to create the subgraph in · smart-piggies
✔ Ethereum network · rinkeby
✔ Contract address · 0x779BEEFFeCD1d26003d1aCE227E2cEe6B41eC54F
✖ Failed to fetch ABI from Etherscan: Unexpected token C in JSON at position 0
```

Jannis:
Fixes #266.

Previously, we would print `Unexpected token C in JSON at position 0`, which
is not very helpful and could confuse users.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -161,7 +161,15 @@ const loadAbiFromEtherscan = async (network, address) =>
         }.etherscan.io/api?module=contract&action=getabi&address=${address}`,
       )
       let json = await result.json()
-      return JSON.parse(json.result)
+
+      // Etherscan returns a JSON object that has a `status`, a `message` and
+      // a `result` field. The `status` is '0' in case of errors and '1' in
+      // case of success
+      if (json.status === '1') {
+        return JSON.parse(json.result)
+      } else {
+        throw new Error('ABI not found, try loading it from a local file')
+      }
     },
   )
 


src/commands/init.js
```diff
@@ -161,7 +161,11 @@ const loadAbiFromEtherscan = async (network, address) =>
         }.etherscan.io/api?module=contract&action=getabi&address=${address}`,
       )
       let json = await result.json()
-      return JSON.parse(json.result)
+      if (json.status === '1' && json.message === 'OK') {
```
leodasvacas: What's the meaning of `status` and `message` here?

Jannis: I couldn't find documentation on it anywhere but I tried valid and invalid requests and the valid ones returned `status: '1', message: 'OK'`, whereas the invalid ones returned `status: '0', message: 'NOK'`.

The only reference I could find was this:
https://github.com/sebs/etherscan-api/blob/242d85a9e5ef066fcabd315dd5c9f95d656845b4/lib/get-request.js#L39-L40

leodasvacas: Got it, could you put that in a comment? Possibly we should check just the status?

Jannis: Done!

leoyvens (approved):

PR #269, opened 2019-05-10
Improve int/uint/float type suggestions in schema validation

### LINKED ISSUE #210: Suggest `Int` when i32 / int32 types etc. are used in the schema

Jannis:
This covers all types from i8..i256, int8..int256, u8..u256 and uint8..uint256
as well as `float` and `Float`.

Resolves #210.


src/validation/schema.js
```diff
@@ -17,21 +17,33 @@ const BUILTIN_SCALAR_TYPES = [
...
+    'BigInt',
+  ],
+]
+
+const typeSuggestion = typeName =>
+  TYPE_SUGGESTIONS.filter(([pattern, _]) => {
+    return typeof pattern === 'string' ? pattern === typeName : typeName.match(pattern)
+  })
+    .map(([_, suggestion]) => suggestion)
+    .find(_ => true)
```
leodasvacas: Is `.find(_ => true)` a way of getting the first item in the iterator? Is there a less hacky way?

leoyvens (commented):
Great! Just one comment.

src/validation/schema.js
Jannis: Done

leoyvens (approved):

PR #270, opened 2019-05-14
Do not add calls with no outputs.

fubhy:
I ran into an issue while trying to run codegen for this contract: 

https://github.com/melonproject/protocol/blob/master/src/contracts/fund/policies/PolicyManager.sol#L67-L73

Now, I am not sure why this is marked public in the first place and I am also not the author of this code but it was causing errors in the code generator due to having 0 outputs.

Thinking about it further, I can think we can safely ignore calls with no outputs :-P.


Jannis:
Great. I agree this makes sense.

Jannis:
Thanks!

PR #273, opened 2019-05-15
Add `log` calls to the example subgraph

Jannis:
This is just to verify that mappings using the `log` API will compile.


lutter (approved):
LGTM

PR #275, opened 2019-05-16
Generate types for Tuple function parameters in codegen

### LINKED ISSUE #272: Tuple fields are not being generated properly
For the following ABI, the tuple fields are not being created, so there is no way to access them:
```
{
   "constant":true,
   "inputs":[
      {
         "name":"proposalId",
         "type":"uint256"
      }
   ],
   "name":"getProposal",
   "outputs":[
      {
         "components":[
            {
               "name":"result",
               "type":"uint8"
            },
            {
               "name":"target",
               "type":"address"
            },
            {
               "name":"data",
               "type":"bytes"
            },
            {
               "name":"proposer",
               "type":"address"
            },
            {
               "name":"feeRecipient",
               "type":"address"
            },
            {
               "name":"fee",
               "type":"uint256"
            },
            {
               "name":"startTime",
               "type":"uint256"
            },
            {
               "name":"yesCount",
               "type":"uint256"
            },
            {
               "name":"noCount",
               "type":"uint256"
            }
         ],
         "name":"",
         "type":"tuple"
      }
   ],
   "payable":false,
   "stateMutability":"view",
   "type":"function"
}
```

The output we are getting is just:
image

It should really look something like:
```class ProposalResultStruct extends EthereumTuple { get target(): Address { return this[0].toAddress(); } etc....```

Interestingly, this ABI is the result we get from the following function:
```
    function getProposal(uint proposalId) external view returns (Proposal memory) {
        return proposals[proposalId];
    }
```
However, when we call the public getter from the below example, it is actually different:
```
    Proposal[] public proposals;
```
ABI:
```
{
   "constant":true,
   "inputs":[
      {
         "name":"",
         "type":"uint256"
      }
   ],
   "name":"proposals",
   "outputs":[
      {
         "name":"result",
         "type":"uint8"
      },
      {
         "name":"target",
         "type":"address"
      },
      {
         "name":"data",
         "type":"bytes"
      },
      {
         "name":"proposer",
         "type":"address"
      },
      {
         "name":"feeRecipient",
         "type":"address"
      },
      {
         "name":"fee",
         "type":"uint256"
      },
      {
         "name":"startTime",
         "type":"uint256"
      },
      {
         "name":"yesCount",
         "type":"uint256"
      },
      {
         "name":"noCount",
         "type":"uint256"
      }
   ],
   "payable":false,
   "stateMutability":"view",
   "type":"function"
}
```

**Solution**
We should be able to handle both of the examples that this is being returned. The second one doesn't actually mention tuple anywhere. Must be a solidity compiler thing. 
The workaround right now for me is that the second example still works with the old way of `proposalData.value0` , etc. 

@Ford

### PR DESCRIPTION
**Pull request by fordN**
Resolves #272 

Add support for Tuple type functions parameters by generating the necessary classes.  Each Tuple type input or output parameter needs it's own class with getters for the mapping handlers to access it's members by name.  

Note: For Tuple type function parameters to be supported by [Graph Node](https://github.com/graphprotocol/graph-node), [PR ethabi#6](https://github.com/graphprotocol/ethabi/pull/6) is required.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -357,6 +357,15 @@ module.exports = class AbiCodeGenerator {
       let returnType = undefined
       let simpleReturnType = true
       let tupleResultParentType = this.abi.name + '__' + fnAlias + 'Result__'
+
+      // Disambiguate outputs with duplicate names
+      let outputs = util.disambiguateNames({
+        values: member.get('outputs', immutable.List()),
+        getName: (input, index) => input.get('name') || `param${index}`,
+        setName: (input, name) => input.set('name', name),
+      })
+      member = member.set('name', fnAlias)
+
       if (member.get('outputs', immutable.List()).size > 1) {
         simpleReturnType = false
 
@@ -369,8 +378,7 @@ module.exports = class AbiCodeGenerator {
         returnType.addMethod(
           tsCodegen.method(
             'constructor',
-            member
-              .get('outputs')
+            outputs
               .map((output, index) =>
                 tsCodegen.param(
                   `value${index}`,
@@ -384,8 +392,7 @@ module.exports = class AbiCodeGenerator {
                 ),
               ),
             null,
-            member
-              .get('outputs')
+            outputs
               .map((output, index) => `this.value${index} = value${index}`)
               .join('\n'),
           ),
@@ -399,8 +406,7 @@ module.exports = class AbiCodeGenerator {
             tsCodegen.namedType('TypedMap<string,EthereumValue>'),
             `
             let map = new TypedMap<string,EthereumValue>();
-            ${member
-              .get('outputs')
+            ${outputs
               .map(
                 (output, index) =>
                   `map.set('value${index}', ${typesCodegen.ethereumValueFromAsc(
@@ -415,8 +421,7 @@ module.exports = class AbiCodeGenerator {
         )
 
         // Add value0, value1 etc. members to the type
-        member
-          .get('outputs')
+        outputs
           .map((output, index) =>
             tsCodegen.klassMember(
               `value${index}`,
@@ -426,7 +431,7 @@ module.exports = class AbiCodeGenerator {
           .forEach(member => returnType.addMember(member))
 
         //Create types for Tuple outputs
-        member.get('outputs').forEach((output, index) => {
+        outputs.forEach((output, index) => {
           if (output.get('type') === 'tuple') {
             types = types.push(
               this._generateTupleType(
@@ -445,15 +450,14 @@ module.exports = class AbiCodeGenerator {
 
         returnType = tsCodegen.namedType(returnType.name)
       } else {
-        let type = member
-          .get('outputs')
+        let type = outputs
           .get(0)
           .get('type')
         if (type === 'tuple') {
           // Add the Tuple type to the types we'll create
           types = types.push(
             this._generateTupleType(
-              member.get('outputs').get(0),
+              outputs.get(0),
               0,
               tupleResultParentType,
               'function',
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/codegen/abi.js [hunk 1]
```diff
@@ -411,37 +430,93 @@ module.exports = class AbiCodeGenerator {
...
+            this._generateTupleType(
+              outputs.get(0),
+              0,
+              tupleResultParentType,
+              'function',
+              this.abi.name,
+            ).classes,
+          )
+
+          returnType = this._generateTupleClassName(
```
Jannis: Is this going to be compatible with the above `_generateTupleType`? Should we maybe add the name of the type to the object returned by `_generateTupleType` and then use that?

src/codegen/abi.js
```diff
@@ -411,37 +430,93 @@ module.exports = class AbiCodeGenerator {
...
       // Disambiguate inputs with duplicate names
       let inputs = util.disambiguateNames({
         values: member.get('inputs', immutable.List()),
         getName: (input, index) => input.get('name') || `param${index}`,
-        setName: (input, name) => input.set('_alias', name),
+        setName: (input, name) => input.set('name', name),
+      })
+
+      // Generate a type suffix to identify the Tuple inputs to a function
+      let tupleInputParentType = this.abi.name + '__' + fnAlias + 'Input__'
```
Jannis: It's a prefix, not a suffix. Also: `Input` instead of `Input__`.

src/codegen/abi.js
```diff
@@ -411,37 +430,93 @@ module.exports = class AbiCodeGenerator {
...
+
+      //Create types for Tuple inputs
```
Jannis: Add space after `//`

src/codegen/abi.js
```diff
@@ -411,37 +430,93 @@ module.exports = class AbiCodeGenerator {
...
         tsCodegen.method(
           fnAlias,
-          inputs.map(input =>
+          inputs.map((input, index) =>
             tsCodegen.param(
-              input.get('_alias'),
-              typesCodegen.ascTypeForEthereum(input.get('type')),
+              input.get('name'),
+              input.get('type') === 'tuple'
+                ? this._generateTupleClassName(
```
Jannis: Same as before: Is this guaranteed to be compatible with the `_generateTupleType` result? I think we should combine the type + name generation.

src/codegen/abi.js
```diff
@@ -466,24 +541,37 @@ module.exports = class AbiCodeGenerator {
...
+                  ${outputs
+                    .map(
+                      (output, index) =>
+                        `${typesCodegen.ethereumValueToAsc(
+                          `result[${index}]`,
+                          output.get('type'),
+                        )} ${
+                          output.get('type') === 'tuple'
+                            ? 'as ' +
+                              this._generateTupleClassName(
```
Jannis: Same here, re: name compatibility.

src/codegen/abi.js
```diff
@@ -239,6 +239,7 @@ module.exports = class AbiCodeGenerator {
           parentField,
         )
       : {
+          name: [],
```
Jannis: What's this for?

fordN: Good catch, the `name` property on this result of this function doesn't get used.  

The other conditional arm here resolves to an object with a `name` property; in order to keep the return type of `_generateInputOrOutput` consistent I can delete `name` wherever it would be returned.

So we'd get something along the lines of
```rust
return valueType === 'tuple'
      ? this._generateTupleType(
          inputOrOutput,
          index,
          parentClass,
          parentType,
          parentField,
        ).delete('name')
      : {
          getter: tsCodegen.method(...
```

Jannis:
@fordN I added a (largely incomplete) suite of tests for ABI code generation and found a couple of bugs (mostly caused by `types = types.push(generateTupleType(...).classes)` being used instead of `types = types.concat(...)`.

This should be ready to go in now.

Jannis:
I'll merge it so we can release a new version.

PR #276, opened 2019-05-16
Allow stateMutability 'nonpayable' in generated functions.

fubhy:
I ran into several contracts and calls that are not explicitly marked as pure/view but do not modify state, e.g.: https://github.com/melonproject/protocol/blob/master/src/contracts/fund/accounting/Accounting.sol#L138

Is there a compelling reason not to generate these functions? Would it break something down the line? Otherwise I think the filter is a bit too restrictive.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -327,13 +327,12 @@ module.exports = class AbiCodeGenerator {
     )
 
     // Get view/pure functions from the contract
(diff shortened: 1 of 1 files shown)


fubhy:
Actually not right. According to my colleague there actually is a state change in that call :-(

Jannis:
@fubhy I'm just seeing your "actually not right" comment. Was that to say we shouldn't generate methods for calling "nonpayable" functions?

PR #280, opened 2019-05-21
Disallow constructor in callHandlers.

fubhy:
Fixes https://github.com/graphprotocol/support/issues/8#issuecomment-492947577

### DIFF
--- a/src/abi.js
+++ b/src/abi.js
@@ -55,16 +55,16 @@ module.exports = class ABI {
   }
 
   callFunctionSignatures() {
-    return this.callFunctions().map(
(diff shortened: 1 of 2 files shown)


fubhy:
I think the tests just need a restart. The failure seems unrelated.

Jannis:
Yep, the `graph auth` test failures are resolved on master.

Jannis:
Yep, the `graph auth` test failures are resolved on master.

Jannis:
We'll want to bring this back in some form later but for now we should remove it, as it isn't yet supported.

PR #281, opened 2019-05-22
Use `indexed` for parameters in event signatures

Jannis:
This changes Graph CLI to distinguish between indexed and non-indexed parameters. Event signatures are now expected to include an `indexed` prefix for indexed parameters. Example:
```
eventHandlers:
  - event: SetOwner(indexed string,uint256)
    ...
```

This change mainly affects subgraph validation, specifically the logic that checks whether an event is present in an ABI.

In addition to this, the PR comes with a migration that automatically updates manifests to use the new signatures, if they are built with Graph CLI > 0.12.0.

Part of resolving https://github.com/graphprotocol/graph-node/issues/913.

### DIFF
--- a/src/migrations/indexed-event-params.js
+++ b/src/migrations/indexed-event-params.js
@@ -1,0 +1,141 @@
+const fs = require('fs-extra')
+const immutable = require('immutable')
+const path = require('path')
+const semver = require('semver')
+const toolbox = require('gluegun/toolbox')
+const yaml = require('yaml')
+
+const ABI = require('../abi')
+const Subgraph = require('../subgraph')
+
+const getGraphCliVersion = sourceDir => {
+  let pkgJsonFile = path.join(
+    sourceDir,
+    'node_modules',
+    '@graphprotocol',
+    'graph-cli',
+    'package.json',
+  )
+  let data = fs.readFileSync(pkgJsonFile)
+  let jsonData = JSON.parse(data)
+  return jsonData.version
+}
+
+const dataSourceHasEventHandlers = dataSource =>
+  dataSource &&
+  typeof dataSource === 'object' &&
+  ((dataSource.mapping &&
+    typeof dataSource.mapping === 'object' &&
+    dataSource.mapping.eventHandlers &&
+    Array.isArray(dataSource.mapping.eventHandlers) &&
+    dataSource.mapping.eventHandlers.length > 0) ||
+    (dataSource.templates &&
+      Array.isArray(dataSource.templates) &&
+      dataSource.templates.reduce(
+        (hasEventHandlers, template) =>
+          hasEventHandlers || dataSourceHasEventHandlers(template),
+        false,
+      )))
+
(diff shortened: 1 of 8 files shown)


Jannis:
The tests are passing, it's just that `dbus-launch` on Travis suddenly has stopped working.

leodasvacas:
Needs a rebase to fix CI.

src/migrations/indexed-event-params.js
```diff
@@ -0,0 +1,141 @@
...
+  if (immutable.List.isList(dataSource.get('templates'))) {
+    dataSource = dataSource.update('templates', templates =>
+      templates.map(template => maybeMigrateDataSource(manifestFile, template)),
+    )
+  }
+
+  return dataSource
+}
+
+// If any of the manifest apiVersions are 0.0.1, replace them with 0.0.2
```
leodasvacas: This comments looks left over from a copy-paste.

src/migrations/indexed-event-params.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    oldEventSignatures = abi.oldEventSignatures()
+    newEventSignatures = abi.eventSignatures()
+  }
+
+  if (oldEventSignatures && newEventSignatures) {
+    for (let i = 0; i < oldEventSignatures.size; i++) {
+      // Only replace the event signature if the signature of the event has changed
+      // and the old signature is not present in the new signatures anymore.
```
leodasvacas: I don't get why this is necessary, why can't we just overwrite all signatures in the new form?

leoyvens (changes requested):
Just some questions on the migration.

src/migrations/indexed-event-params.js
Jannis: It's not super obvious.

Let's say an event has more than one variant, e.g. one without indexed params and one with indexed params, e.g.
```yaml
SomeEvent(uint256)
SomeEvent(indexed uint256)
```

In the old manifest we'll find `SomeEvent(uint256)`. The question then is: do we translate this to `SomeEvent(indexed uint256)` or do we keep `SomeEvent(uint256)`. Both are present in the new set of signatures.

If `SomeEvent(uint256)` is no longer present, we can simply replace it. But if it's still around, this is a decision we can't really make on the users behalf because they could mean either of the variants. I feel in this case it's best to leave the signature untouched.

Jannis: I'll update the comment to include this explanation.

src/migrations/indexed-event-params.js
Jannis: Indeed

Jannis:
@leodasvacas Comments addressed!

src/migrations/indexed-event-params.js
leodasvacas: Great explanation, thanks!

leoyvens (approved):

PR #282, opened 2019-05-24
Validate @derivedFrom directives

### LINKED ISSUE #279: @derivedFrom specifying a nonexistent field does not raise an error
The following schema should raise an error:

```
type Foo @entity {
  id: ID!
  baz: Bar
}

type Bar @entity {
  id: ID!
  foos: [Foo!]! @derivedFrom(field: "bar")
}
```

However, it compiles just fine, and returns no results for `foos` at runtime.

Jannis:
Fixes #279 by adding validation rules that catch

- `@derivedFrom` directives that don't have a `field` argument
- `@derivedFrom` directives that have non-string `field` values
- `@derivedFrom` directives whose `field` doesn't exist in the target entity
- `@derivedFrom` directives whose `field` doesn't refer back to the source entity

### DIFF
--- a/src/validation/schema.js
+++ b/src/validation/schema.js
@@ -135,7 +135,7 @@ Reason: Lists with null elements are not supported.`,
         ])
       : List()
 
-const validateInnerFieldType = (defs, def, field) => {
+const unwrapType = type => {
   let innerTypeFromList = listType =>
     listType.type.kind === 'NonNullType'
       ? innerTypeFromNonNull(listType.type)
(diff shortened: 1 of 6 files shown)


leodasvacas:
Needs a rebase to fix CI.

Jannis:
@leodasvacas Done!

leoyvens (approved):
Nice! Custom error messages are always nice, but the missing field check and the type check would be caught by regular graphql schema validation, right? Do we do that?

Jannis:
I don't think we do right now.

PR #285, opened 2019-06-10
Add mapping_api_version_0_0_2 migration

olibearo:
https://github.com/graphprotocol/graph-node/pull/976
https://github.com/graphprotocol/graph-ts/pull/66/

### DIFF
--- a/src/migrations/mapping_api_version_0_0_1.js
+++ b/src/migrations/mapping_api_version_0_0_1.js
@@ -2,7 +2,7 @@ const fs = require('fs-extra')
 const semver = require('semver')
 const toolbox = require('gluegun/toolbox')
 const yaml = require('js-yaml')
-const getGraphTsVersion = require('./util/versions').getGraphTsVersion
+const { getGraphTsVersion } = require('./util/versions')
 
 // If any of the manifest apiVersions are 0.0.1, replace them with 0.0.2
 module.exports = {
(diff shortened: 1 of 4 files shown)


src/migrations/mapping_api_version_0_0_2.js
```diff
@@ -0,0 +1,57 @@
...
+    } catch (_) {
+      // If we cannot obtain the version, return a hint that the graph-ts
+      // hasn't been installed yet
+      return 'graph-ts dependency not installed yet'
+    }
+
+    let manifest = yaml.safeLoad(fs.readFileSync(manifestFile, 'utf-8'))
+    return (
+      // Only migrate if the graph-ts version is > 0.5.1...
+      semver.gt(graphTsVersion, '0.5.1') &&
```
Jannis: The current version is 0.12.0, so this should be `0.12.0`, so it only takes effect starting with the next graph-ts release we make.

src/migrations/mapping_api_version_0_0_1.js
```diff
@@ -1,21 +1,8 @@
...
-    'node_modules',
-    '@graphprotocol',
-    'graph-ts',
-    'package.json',
-  )
-  let data = fs.readFileSync(pkgJsonFile)
-  let jsonData = JSON.parse(data)
-  return jsonData.version
-}
+const getGraphTsVersion = require('./util/versions').getGraphTsVersion
```
Jannis: I'd do this here (and in the other migration):
```
const { getGraphTsVersion } = require('./util/versions`)
```

PR #288, opened 2019-06-13
Sequential migrations, remove indexed event params migration

Jannis:
Until now, subgraph migrations were performed in parallel, which is _not_ a good idea. This PR changes it so they are executed one after the other, in the desired order.

This PR also removes the migration to replace event signatures with `indexed` event params. The migration unintentionally removed comments, messed with the order of fields in the manifest and other formatting. That's simply too invasive.


leoyvens (approved):
I couldn't find a JS yaml parser that retained comments and formatting.

Jannis:
I tried https://www.npmjs.com/package/yawn-yaml and changed fewer things in the YAML but still erased all comments, despite claiming not to do that.

PR #289, opened 2019-06-14
Generate code for functions that use deprecated JSON fields

Jannis:
Specifically, add support for generating callable methods for functions that don't use `stateMutability` and use the deprecated `payable`/`constant` fields instead.

This fix includes updated tests to prevent regressions.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -334,7 +334,8 @@ module.exports = class AbiCodeGenerator {
       member =>
         member.get('type') === 'function' &&
         member.get('outputs', immutable.List()).size !== 0 &&
-        allowedMutability.indexOf(member.get('stateMutability')) !== -1,
+        (allowedMutability.indexOf(member.get('stateMutability')) >= 0 ||
+          (member.get('payable') === false && member.get('constant') === true)),
     )
 
     // Disambiguate functions with duplicate names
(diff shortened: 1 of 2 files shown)


Jannis:
From the Solidity docs:

> `type` can be omitted, defaulting to "function", likewise `payable` and `constant` can be omitted, both defaulting to `false`.

Perhaps we can even support any function that has `payable` set to `false`?

src/codegen/abi.js
```diff
@@ -334,7 +334,8 @@ module.exports = class AbiCodeGenerator {
       member =>
         member.get('type') === 'function' &&
         member.get('outputs', immutable.List()).size !== 0 &&
-        allowedMutability.indexOf(member.get('stateMutability')) !== -1,
+        (allowedMutability.indexOf(member.get('stateMutability')) >= 0 ||
```
leodasvacas: Isn't [includes](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes) more elegant than `indexOf`?

leoyvens (changes requested):

src/codegen/abi.js
Jannis: You're right, I've updated the check.

Jannis:
I've changed the code not to check for `constant` (because we don't care) and to generated code for any function that either

* has one of the supported `stateMutability` values, or
* doesn't have `stateMutability` and either has `payable` set to `false` or no `payable` value at all (it defaults to `false`).

leoyvens (approved):

PR #293, opened 2019-07-02
Fix string array conversions in generated ABI code

Jannis:
This fixes a bug where `graph init` would fail to scaffold a subgraph
that was using `string[]` in the ABI.


leodasvacas:
Is it possible to get a test for this?

Jannis:
@leodasvacas I've added some tests for type conversions/relationships.

leodasvacas:
Nice!

leoyvens (approved):

PR #312, opened 2019-07-17
Make `--from-contract` generate a more instructive scaffold

Jannis:
This changes the default scaffold create by `graph init --from-contract` to

1. only generate a single example entity rather than entities for all events
2. generate empty event handlers (apart from one)
3. add one event handler that includes an example of how to work with entities as well as comments about binding contracts and making calls.

Generating a subgraph that indexes events 1:1 is still possible by using the `--index-events` flag.

As a little extra feature, this PR also adds support for creating subgraphs from contracts on the Goerli testnet.

The two sections below show what is generated for `0xF87E31492Faf9A91B02Ee0dEAAd50d51d56D5d4d` on `mainnet`.

## Generated schema

The example entity has two fixed fields (`id` and `count`, to show how an existing entity can be updated) and up to two fields taken from the first event type we can find (to show how event parameters can be passed on to entities).

``` graphql
type ExampleEntity @entity {
  id: ID!
  count: BigInt!
  newContract: Bytes! # address
  initializedWith: Bytes! # bytes
}
```

## Generated mapping

``` typescript
import { BigInt } from "@graphprotocol/graph-ts"
import { Contract, Upgrade, OwnerUpdate } from "../generated/Contract/Contract"
import { ExampleEntity } from "../generated/schema"

export function handleUpgrade(event: Upgrade): void {
  // Entities can be loaded from the store using an event ID (a string)
  let entity = ExampleEntity.load(event.transaction.from.toHex())

  // Entities only exist after they have been saved to the store;
  // `null` checks allow to create entities on demand
  if (entity == null) {
    entity = new ExampleEntity(event.transaction.from.toHex())

    // Entity fields can be set using simple assignments
    entity.count = BigInt.fromI32(0)
  }

  // BigInt and BigDecimal math are supported
  entity.count = entity.count + BigInt.fromI32(1)

  // Entity fields can be set based on event parameters
  entity.newContract = event.params.newContract
  entity.initializedWith = event.params.initializedWith

  // Entities can be written to the store with `.save()`
  entity.save()

  // Note: If a handler doesn't require existing field values, it is faster
  // _not_ to load the entity from the store. Instead, create it fresh with
  // `new Entity(...)`, set the fields that should be updated and save the
  // entity back to the store. Fields that were not set or unset remain
  // unchanged, allowing for partial updates to be applied.

  // It is also possible to access smart contracts from mappings. For
  // example, the contract that has emitted the event can be connected to
  // with:
  //
  // let contract = Contract.bind(event.address)
  //
  // The following functions can then be called on this contract to access
  // state variables and other data:
  //
  // - contract.proxyOwner(...)
  // - contract.currentContract(...)
  // - contract.owner(...)
}

export function handleOwnerUpdate(event: OwnerUpdate): void {}
```

### DIFF
--- a/src/abi.js
+++ b/src/abi.js
@@ -5,19 +5,21 @@ const path = require('path')
 const AbiCodeGenerator = require('./codegen/abi')
 
 const buildOldSignatureParameter = input => {
-  return input.type === 'tuple'
-    ? `(${input.components
+  return input.get('type') === 'tuple'
+    ? `(${input
+        .get('components')
         .map(component => buildSignatureParameter(component))
         .join(',')})`
-    : `${input.type}`
+    : `${input.get('type')}`
 }
 
 const buildSignatureParameter = input => {
-  return input.type === 'tuple'
-    ? `(${input.indexed ? 'indexed ' : ''}${input.components
+  return input.get('type') === 'tuple'
+    ? `(${input.get('indexed') ? 'indexed ' : ''}${input
+        .get('components')
         .map(component => buildSignatureParameter(component))
         .join(',')})`
-    : `${input.indexed ? 'indexed ' : ''}${input.type}`
+    : `${input.get('indexed') ? 'indexed ' : ''}${input.get('type')}`
 }
 
 module.exports = class ABI {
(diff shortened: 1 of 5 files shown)


leoyvens (approved):
Tried locally, worked fine, I think the new content will do a better job at onboarding through an example. Code looks good. We might want more people to review the content, but approved by me.

leodasvacas:
I do have one comment on: `// Entities can be loaded from the store using an event ID (a string)`. I don't think this is a good explanation of what IDs are, it make seem as if the entity ID must be the "event ID" which isn't even a thing.

Jannis:
@leodasvacas Agreed, I'll update!

PR #320, opened 2019-08-01
Add support for Tuple arrays (ABIEncoderV2)

### LINKED ISSUE #316: Add codegen for array of tuples types (ABIEncoderV2)
Generate classes with conversions between `EthereumValue` and `AssemblyScript` for tuple arrays (`tuple[]`) found in the ABI.

fordN:
Resolves #316 
Requires [graph-ts #72](https://github.com/graphprotocol/graph-ts/pull/73)

This PR adds support for generating types from Tuple array type params in the ABI.  The conversion methods from a Tuple array `EthereumValue` is the first conversion method we have that uses a generic type, so it requires special casing in `ethereumValueToAsc`. 

----
_Note: This PR is currently in the draft state because it is has not completed sufficient testing_
It is being tested with subgraphs that have an ABI with a Tuple array type as a:
  - [x] call method param 
  - [x] contract method param
  - [x] event param

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -377,7 +377,7 @@ module.exports = class AbiCodeGenerator {
                       this.abi.name,
                       '',
                     ).name
-                  : output.get('type').substring(0, 6) === 'tuple['
+                  : util.isTupleArrayType(output.get('type'))
                   ? `Array<${
                       this._generateTupleType(
                         output,
@@ -431,7 +431,7 @@ module.exports = class AbiCodeGenerator {
 
         // Create types for Tuple outputs
         outputs.forEach((output, index) => {
-          if (output.get('type').substring(0, 5) === 'tuple') {
+          if (util.containsTupleType(output.get('type'))) {
             types = types.concat(
               this._generateTupleType(
                 output,
@@ -450,7 +450,7 @@ module.exports = class AbiCodeGenerator {
         returnType = tsCodegen.namedType(returnType.name)
       } else {
         let type = outputs.get(0).get('type')
-        if (type.substring(0, 5) === 'tuple') {
+        if (util.containsTupleType(type)) {
           // Add the Tuple type to the types we'll create
           let tuple = this._generateTupleType(
             outputs.get(0),
@@ -481,7 +481,7 @@ module.exports = class AbiCodeGenerator {
 
       // Create types for Tuple inputs
       inputs.forEach((input, index) => {
-        if (input.get('type').substring(0, 5) === 'tuple') {
+        if (util.containsTupleType(input.get('type'))) {
           types = types.concat(
             this._generateTupleType(
               input,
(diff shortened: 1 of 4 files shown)


Jannis:
@fordN I assume the draft state note and the incomplete check boxes in the PR description are no longer up to date?

src/codegen/abi.js
```diff
@@ -520,7 +535,19 @@ module.exports = class AbiCodeGenerator {
           );
           return ${
             simpleReturnType
-              ? typesCodegen.ethereumValueToAsc('result[0]', outputs.get(0).get('type'))
+              ? typesCodegen.ethereumValueToAsc(
+                'result[0]', 
+                outputs.get(0).get('type'), 
+                outputs.get(0).get('type')==='tuple[]'
```
Jannis: The formatting is off here (no spaces before/after the `===`). Make sure you run prettier against the changes.

src/codegen/types/conversions.js
```diff
@@ -44,9 +44,10 @@ const ETHEREUM_VALUE_TO_ASSEMBLYSCRIPT = [
   ],
   [/^string\[([0-9]+)?\]$/, 'Array<string>', code => `${code}.toStringArray()`],
 
-  // Tuple
+  // Tuples
 
   ['tuple', 'EthereumTuple', code => `${code}.toTuple()`],
+  [/^tuple\[([0-9]+)?\]$/, 'Array<EthereumTuple>', (code, type) => `${code}.toTupleArray<${type}>()`],
```
Jannis: So far we've been keeping the type casts out of this (see `toTuple()` above) and added ` as ${type}` in the code generation. I think I prefer that slightly.

Jannis: After looking at https://github.com/graphprotocol/graph-ts/pull/73 I see why you did this. Let's keep it!

src/codegen/abi.js
```diff
@@ -377,6 +377,14 @@ module.exports = class AbiCodeGenerator {
                       this.abi.name,
                       '',
                     ).name
+                  : output.get('type') === 'tuple[]'
```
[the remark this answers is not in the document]
fordN: Yes of course, there also could be fixed size tuple arrays in the ABI.   
I've updated the tuple array type codegen to also handle fixed size arrays, i.e `tuple[5]`.

fordN:
@Jannis I've updated the PR to handle fixed sized tuple arrays and ran prettier to ensure canonical formatting.

src/codegen/abi.js
```diff
@@ -377,6 +377,16 @@ module.exports = class AbiCodeGenerator {
                       this.abi.name,
                       '',
                     ).name
+                  : output.get('type').substring(0, 6) === 'tuple['
```
Jannis: I'm not a big fan of matching only the beginning of the string. I'd rather do
```
output.get('type').match(/^tuple\[([0-9]+)?\]$/)
```
Same in the other places. This could be moved into an `isTupleArrayType(s)` helper.

PR #332, opened 2019-08-28
Expose reverts to mappings

leoyvens:
Part of https://github.com/graphprotocol/graph-node/issues/1120.

In addition to the contract calls we already generate, we'll generate a version prefixed with `try_` that returns a `CallResult<T>` allowing the mapping to handle failures. See graph-ts PR https://github.com/graphprotocol/graph-ts/pull/74.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -523,31 +523,35 @@ module.exports = class AbiCodeGenerator {
 
       let methodCallBody = isTry =>
         `
-        let result = ${isTry ? 'super.tryCall' : 'super.call'}(${superInputs})
         ${
           isTry
-            ? `if (result.reverted) {
-            return new CallResult()
-          }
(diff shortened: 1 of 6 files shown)


src/codegen/abi.js
```diff
@@ -496,83 +497,91 @@ module.exports = class AbiCodeGenerator {
...
-                                index,
-                                tupleResultParentType,
-                                this.abi.name,
-                                '',
-                              ).name
-                            : ''
-                        }`,
+            : ''
+        }
+        return ${isTry ? 'CallResult.fromValue(' : ''} ${
```
Jannis: This section might be easier to read if there was just one top-level `isTry`-based conditional rather than a few spread across the nested logic.

leoyvens:
@Jannis you're right, we can collapse the conditionals, thanks!

PR #337, opened 2019-09-02
Migration fixes, IPFS upload optimization

Jannis:
This PR fixes a couple of things:

1. Existing `apiVersion` migrations are updated to replace _all_ `apiVersion` occurrences. Before, only the first would be replaced.

2. `--skip-migrations` is now properly passed on to the compiler. For some time, it wasn't.

These fixes are necessary for https://github.com/daostack/subgraph/pull/318 to work.

Another change made in this PR is that during a `graph build --ipfs ...` and `graph deploy`, files are only uploaded to IPFS once. Any repeated upload attempts simply pick the IPFS hash from a short-lived cache. This speeds up build/deploy for subgraphs that reuse ABIs or mappings across many data sources (like DAOstack).

### DIFF
--- a/src/command-helpers/compiler.js
+++ b/src/command-helpers/compiler.js
@@ -3,7 +3,7 @@ const toolbox = require('gluegun/toolbox')
 const Compiler = require('../compiler')
 
 // Helper function to construct a subgraph compiler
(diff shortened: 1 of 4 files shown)


Jannis:
I'm going to take the liberty and merge this.

PR #352, opened 2019-09-25
Check for duplicate data source and template names

### LINKED ISSUE #326: Validate against data sources with the same name
Otherwise they'll overwrite each others `.wasm` and silently cause a mess.

leoyvens:
Resolves #326.


leoyvens:
@Jannis this wants a review.

PR #360, opened 2019-10-01
Do not assume the source dir is the working dir

### LINKED ISSUE #359: Graph CLI to accept argument to path of subgraph.yaml
We use the same subgraph repo for a number of similar subgraphs. Yet the CLI is limited in that it expects a file `./subgraph.yaml` in the root of the repo. In order to support our workflow, we'd like the path to the `yaml` file to be optionally provided to `graph-cli` as an argument. This applies to the `build`, `codegen` and `deploy` actions. 

This is how we're currently working around the limitation. https://github.com/Synthetixio/synthetix-subgraph/blob/master/package.json#L10-L13

leoyvens:
We were doing that in some places. Resolves #359.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -52,6 +52,7 @@ class Compiler {
         return true
       }
     } catch (e) {
+      console.log(e)
       return false
     }
   }
@@ -208,7 +209,8 @@ class Compiler {
       )
 
       let baseDir = this.sourceDir
-      let inputFile = path.relative(baseDir, mappingPath)
+      let absoluteMappingPath = path.resolve(baseDir, mappingPath)
+      let inputFile = path.relative(baseDir, absoluteMappingPath)
       let outputFile = path.relative(baseDir, outFile)
 
       // Create output directory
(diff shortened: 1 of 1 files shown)


src/compiler.js
```diff
@@ -52,6 +52,7 @@ class Compiler {
         return true
       }
     } catch (e) {
+      console.log(e)
```
Jannis: Do we want this log message? _If_ we log anything here, it should be done using the `print` module of the Gluegun toolbox.

Jannis (changes requested):
One minor remark about `console.log`, everything else looks good to me. I confirmed the bug on master.

src/compiler.js
leoyvens: I think usually this should never happen, I hit it due to a double variable declaration while developing. Changed this to use gluegun.

Jannis (approved):

src/compiler.js
Jannis: Ah, yeah, any errors during compilation should already be logged inside the `try {}` branch.

PR #365, opened 2019-10-07
Check directories above for libs

### LINKED ISSUE #359: Graph CLI to accept argument to path of subgraph.yaml

leoyvens:
To fix #359 we still needed to be able to find `node_modules` based on the `subgraph.yaml` path. This makes it so that we also check a few directories above to find `node_modules`, there may still be cases this doesn't cover, but it covers the known use case.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -234,7 +234,9 @@ class Compiler {
       }
 
       if (libs === undefined) {
-        throw Error(`could not find \`node_modules\` path`)
+        throw Error(
+          `could not locate \`node_modules\` in parent directories of subgraph manifest`,
+        )
       }
 
       let global = path.join(libs, '@graphprotocol', 'graph-ts', 'global', 'global.ts')
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -220,18 +220,23 @@ class Compiler {
         throw e
       }
 
-      // Look for `node_modules` in the same dir as `subgraph.yaml`.
-      let libs = path.join(baseDir, 'node_modules')
-      if (!fs.existsSync(libs)) {
-        // It's not there, so look in the directories above. This, in practice, checks all ancestor
-        // directories but keeps the loop simple.
-        for (let i = 1; i <= 100; i++) {
-          let upwardPath = path.resolve(libs, ...Array(i).fill('..'), 'node_modules')
-          if (fs.existsSync(upwardPath)) {
-            libs = upwardPath
-          }
+      let libs
+      for (
+        let dir = path.resolve(baseDir);
+        // Terminate after the root dir or when we have found node_modules
+        dir !== undefined && libs === undefined;
+        // Continue with the parent directory, terminate after the root dir
+        dir = path.dirname(dir) === dir ? undefined : path.dirname(dir)
+      ) {
+        if (fs.existsSync(path.join(dir, 'node_modules'))) {
+          libs = path.join(dir, 'node_modules')
         }
       }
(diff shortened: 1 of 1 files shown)


src/compiler.js
```diff
@@ -220,7 +220,17 @@ class Compiler {
         throw e
       }
 
+      // Look for `node_modules` in the same dir as `subgraph.yaml`.
       let libs = path.join(baseDir, 'node_modules')
+      if (!fs.existsSync(libs)) {
+        // It's not there, so look in up to three directories above.
+        for (let i = 1; i <= 3; i++) {
```
Jannis: I'd check _all_ parent directories. I know `/` doesn't make sense to check but directories are rarely deeper than, say 10 levels, so this check would be cheap to run for all of them, and special-casing to 3 feels unnecessary.

Jannis (changes requested):

src/compiler.js
leoyvens: I took the lazy/simple route and bumped 3 to 100, so we don't have to use a complicated logic that makes it hard to tell if it can loop forever.

src/compiler.js
```diff
@@ -220,7 +220,18 @@ class Compiler {
...
+      if (!fs.existsSync(libs)) {
+        // It's not there, so look in the directories above. This, in practice, checks all ancestor
+        // directories but keeps the loop simple.
+        for (let i = 1; i <= 100; i++) {
+          let upwardPath = path.resolve(libs, ...Array(i).fill('..'), 'node_modules')
+          if (fs.existsSync(upwardPath)) {
+            libs = upwardPath
+          }
+        }
+      }
```
Jannis: Rather than the slightly obscure iteration over `i`, wouldn't something like this be better (and more obvious)?
```
const fs = require('fs')
const path = require('path')

let manifestFile = '/usr/local/share/subgraph.yaml'

let libs
for (
  let dir = baseDir;
  libs === undefined && dir !== undefined;
  dir = path.dirname(dir) === dir ? undefined : path.dirname(dir)
) {
  if (fs.existsSync(path.join(dir, 'local'))) {
    libs = dir
  }
}
```
You can play with it here: https://repl.it/repls/MintyStripedLinuxpc

src/compiler.js
```diff
@@ -220,7 +220,23 @@ class Compiler {
...
+        // Continue with the parent directory, terminate after the root dir
+        dir = path.dirname(dir) === dir ? undefined : path.dirname(dir)
+      ) {
+        if (fs.existsSync(path.join(dir, 'node_modules'))) {
+          libs = path.join(dir, 'node_modules')
+        }
+      }
+
+      if (libs === undefined) {
+        throw Error(`could not find \`node_modules\` path`)
```
Jannis: Let's make this a little bit more helpful by saying
```
could not locate `node_modules` in parent directories of subgraph manifest `{}`
```

leoyvens: Thanks, that's better.

Jannis (approved):

PR #368, opened 2019-10-16
Support an optional startBlock field on data sources

fordN:
Required to leverage the start block data source filtering during syncing as proposed in `graph-node PR#1262`.

This PR updates manifest validation to process BigInts from the manifest and adds the `startBlock` field to the `EthereumContractSource` type.

### DIFF
--- a/src/validation/manifest.js
+++ b/src/validation/manifest.js
@@ -147,6 +147,16 @@ const validators = immutable.fromJS({
           },
         ]),
 
+  BigInt: (value, ctx) =>
+    typeof value === 'number'
+      ? List()
+      : immutable.fromJS([
+        {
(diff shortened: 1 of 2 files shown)


src/validation/manifest.js
```diff
@@ -147,6 +147,16 @@ const validators = immutable.fromJS({
           },
         ]),
 
+  BigInt: (value, ctx) =>
+    typeof value === 'number'
+      ? List()
+      : immutable.fromJS([
+        {
+          path: ctx.get('path'),
+          message: `Expected bigint, found ${typeName(value)}:\n${toYAML(value)}`,
```
Jannis: `bigint` -> `BigInt`

PR #392, opened 2019-11-12
Subgraph composition

olibearo:
This PR adds support for `@imports` directives which allow subgraphs to imports types from other subgraphs. 

Ideally `@import` directives would be available at the top level of a schema document, but top level directives are not yet available. Relevant issues:

https://github.com/graphql/graphql-spec/issues/410
https://github.com/graphql/graphql-spec/pull/556
https://github.com/Urigo/graphql-import/issues/89

As a convention, the type `_SubgraphSchema_` is reserved exclusively to define `@import` directives. Once top level directives are supported in GraphQL, we can remove this convention and provide a migration script for `_SubgraphSchema_` defined imports.

### DIFF
--- a/src/validation/schema.js
+++ b/src/validation/schema.js
@@ -154,6 +154,65 @@ const unwrapType = type => {
     : type
 }
 
+const gatherLocalTypes = defs =>
+  defs
+    .filter(
+      def =>
+        def.kind === 'ObjectTypeDefinition' ||
+        def.kind === 'EnumTypeDefinition' ||
+        def.kind === 'InterfaceTypeDefinition',
+    )
+    .map(def => def.name.value)
+
+const gatherForeignTypes = defs =>
+  defs
+    .filter(
+      def =>
+        def.kind === 'ObjectTypeDefinition' &&
+        def.name.value == '_SubgraphSchema_' &&
+        def.directives.find(
+          directive =>
+            directive.name.value == 'imports' &&
+            directive.arguments.find(argument => argument.name.value == 'types'),
+        ),
+    )
+    .map(def =>
+      def.directives
+        .filter(
+          directive =>
+            directive.name.value == 'imports' &&
+            directive.arguments.find(argument => argument.name.value == 'types'),
+        )
+        .map(imp =>
+          imp.arguments.find(
+            argument =>
+              argument.name.value == 'types' && argument.value.kind == 'ListValue',
+          ),
+        )
+        .map(arg =>
+          arg.value.values.map(type =>
+            type.kind == 'StringValue'
+              ? type.value
+              : type.kind == 'ObjectValue' &&
+                type.fields.find(
+                  field => field.name.value == 'as' && field.value.kind == 'StringValue',
+                )
+              ? type.fields.find(field => field.name.value == 'as').value.value
+              : undefined,
+          ),
+        ),
+    )
+    .reduce(
+      (flattened, types_args) =>
+        flattened.concat(
+          types_args.reduce((flattened, types_arg) => {
+            types_arg.forEach(type => (type ? flattened.push(type) : undefined))
+            return flattened
+          }, []),
+        ),
+      [],
+    )
+
 const entityTypeByName = (defs, name) =>
   defs
     .filter(
@@ -180,56 +239,8 @@ const validateInnerFieldType = (defs, def, field) => {
   // Collect all types that we can use here: built-ins + entities + enums + interfaces
   let availableTypes = List.of(
     ...BUILTIN_SCALAR_TYPES,
-    ...defs
-      .filter(
-        def =>
-          def.kind === 'ObjectTypeDefinition' ||
-          def.kind === 'EnumTypeDefinition' ||
-          def.kind === 'InterfaceTypeDefinition',
-      )
-      .map(def => def.name.value),
-    ...defs
-      .filter(
-        def =>
-          def.kind === 'ObjectTypeDefinition' &&
-          def.name.value == '_SubgraphSchema_' &&
-          def.directives.find(
-            directive =>
-              directive.name.value == 'imports' &&
-              directive.arguments.find(argument => argument.name.value == 'types'),
-          ),
-      )
-      .map(def => {
-        let imports = def.directives.filter(
-          directive =>
-            directive.name.value == 'imports' &&
-            directive.arguments.find(argument => argument.name.value == 'types'),
-        )
-        let type_args = imports.map(imp =>
-          imp.arguments.find(
-            argument =>
-              argument.name.value == 'types' && argument.value.kind == 'ListValue',
-          ),
-        )
-        return type_args.map(arg =>
-          arg.value.values.map(type =>
-            type.kind == 'StringValue'
-              ? type.value
-              : type.kind == 'ObjectValue' &&
-                type.fields.find(
-                  field => field.name.value == 'as' && field.value.kind == 'StringValue',
-                )
-              ? type.fields.find(field => field.name.value == 'as').value.value
-              : undefined,
-          ),
-        )
-      })
-      .reduce((flattened, types_args) => {
-        return types_args.reduce((flattened, types_arg) => {
-          types_arg.forEach(type => (type ? flattened.push(type) : undefined))
-          return flattened
-        }, [])
-      }, []),
+    ...gatherLocalTypes(defs),
+    ...gatherForeignTypes(defs),
   )
 
   // Check whether the type name is available, otherwise return an error
--- a/src/validation/schema.js
+++ b/src/validation/schema.js
@@ -504,6 +504,8 @@ const validateTypeHasNoFields = def => {
   return errors
 }
 
+const validateAtLeastOneExtensionField = def => List()
+
 const typeDefinitionValidators = {
   ObjectTypeDefinition: (defs, def) =>
     def.name && def.name.value == '_SubgraphSchema_'
@@ -514,7 +516,7 @@ const typeDefinitionValidators = {
           ...validateEntityFields(defs, def),
           ...validateNoImportsDirective(def),
         ),
-  ObjectTypeExtension: (_defs, _def) => List(),
+  ObjectTypeExtension: (_defs, def) => List.of(...validateAtLeastOneExtensionField(def)),
 }
 
 const validateTypeDefinition = (defs, def) =>
--- a/src/validation/schema.js
+++ b/src/validation/schema.js
@@ -322,13 +322,196 @@ const validateEntityFields = (defs, def) =>
     List(),
   )
 
+const validateNoImportsDirective = def =>
+  def.directives.find(directive => directive.name.value == 'imports')
+    ? List().push(
+        immutable.fromJS({
+          loc: def.name.loc,
+          entity: def.name.value,
+          message: '@imports directive only allowed on `_SubgraphSchema_` type',
+        }),
+      )
+    : List()
+
+const importDirectiveTypeValidators = {
+  StringValue: (_def, _directive, _type) => List(),
+  ObjectValue: (def, directive, type) => {
+    let errors = List()
+    if (type.fields.length != 2) {
+      return errors.push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message: '@imports type objects accept and require two fields: [name, as]',
+        }),
+      )
+    }
+    return type.fields.reduce((errors, field) => {
+      if (!['name', 'as'].includes(field.name.value)) {
+        return errors.push(
+          immutable.fromJS({
+            loc: directive.name.loc,
+            entity: def.name.value,
+            message: `@imports type object field '${field.name.value}' invalid,  may only be one of: [name, as]`,
+          }),
+        )
+      }
+      if (field.value.kind != 'StringValue') {
+        return errors.push(
+          immutable.fromJS({
+            loc: directive.name.loc,
+            entity: def.name.value,
+            message: '@imports type object fields [name, as] must be strings',
+          }),
+        )
+      }
+      return errors
+    }, errors)
+  },
+}
+
+const validateImportDirectiveType = (def, directive, type) =>
+  importDirectiveTypeValidators[type.kind]
+    ? importDirectiveTypeValidators[type.kind(def, directive, type)]
+    : List().push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message: 'Imported type must be either String | {name: String, as: String}',
+        }),
+      )
+
+const validateImportDirectiveArgumentTypes = (def, directive, argument) => {
+  if (argument.value.kind != 'ListValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: types must be an list',
+      }),
+    )
+  }
+
+  return argument.value.values.reduce(
+    (errors, type) => errors.append(validateImportDirectiveType(def, directive, type)),
+    List(),
+  )
+}
+
+const validateImportDirectiveArgumentFrom = (def, directive, argument) => {
+  if (argument.value.kind != 'ObjectValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: `from` must be an object',
+      }),
+    )
+  }
+  if (argument.value.fields.length != 1) {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: `from` must have one field of [id, name]',
+      }),
+    )
+  }
+  return argument.value.fields.reduce((errors, field) => {
+    if (!['name', 'id'].includes(field.name.value)) {
+      return errors.push(
+        immutable.fromJS({
+          loc: field.name.loc,
+          entity: def.name.value,
+          message: '@imports directive from: only fields `id` or `name` allowed',
+        }),
+      )
+    }
+    if (field.value.kind != 'StringValue') {
+      return errors.push(
+        immutable.fromJS({
+          loc: field.name.loc,
+          entity: def.name.value,
+          message: '@imports directive from: fields must have string values',
+        }),
+      )
+    }
+  })
+}
+
+const validateImportDirective = (def, directive) => {
+  let errors = List()
+
+  if (directive.name.value != 'imports') {
+    return errors.push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '_SubgraphSchema_ directives: only @import directives allowed',
+      }),
+    )
+  }
+
+  let types = directive.arguments.find(argument => argument.name.value == 'types')
+  if (!types) {
+    errors = errors.push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: `@import argument 'types' must be specified`,
+      }),
+    )
+  } else {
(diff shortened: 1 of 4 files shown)


src/validation/schema.js
```diff
@@ -322,13 +375,201 @@ const validateEntityFields = (defs, def) =>
...
+const validateAtLeastOneExtensionField = def => List()
+
 const typeDefinitionValidators = {
   ObjectTypeDefinition: (defs, def) =>
-    List.of(
-      ...validateEntityDirective(def),
-      ...validateEntityID(def),
-      ...validateEntityFields(defs, def),
-    ),
+    def.name && def.name.value == '_SubgraphSchema_'
```
olibearo: Add a constant for the `_SubgraphSchema_` string

src/validation/schema.js
```diff
@@ -339,10 +583,36 @@ const validateTypeDefinition = (defs, def) =>
 const validateTypeDefinitions = defs =>
   defs.reduce((errors, def) => errors.concat(validateTypeDefinition(defs, def)), List())
 
+const validateNamingCollisionsInTypes = types => {
+  let memo = Set()
```
olibearo: Add a Set to avoid erroring more than once for a type defined > 2 times

src/validation/schema.js
```diff
@@ -38,6 +38,9 @@ const TYPE_SUGGESTIONS = [
   ],
 ]
 
+// As a convention, the type _SubgraphSchema_ is reserved to define imports on.
+const SUBGRAPH_SCHEMA = '_SubgraphSchema_'
```
Jannis: Not sure I like this name yet. The `Subgraph` prefix feels unnecessary. I'd almost want to make it something ridiculous like
```graphql
type _ @import(....) { ... }
```
One concern with using a `type` to add `@imports` to that I'm only realizing now is that the
```graphql
type _SubgraphSchema_ { ... }
```
definition needs to contain at least one field to be valid. We wouldn't want users to have to define anything inside the type though.

I now realize we can still make the
```graphql
# @import(...)
```
trick that I had initially planned work without messing with library internals. We could just take the schema file, extract the first comment and put `type _ .... { id: ID! }` around it. Then parse that the usual way and pull the directive information out of the resulting AST.

src/validation/schema.js
```diff
@@ -339,10 +582,39 @@ const validateTypeDefinition = (defs, def) =>
...
+      errored = errored.add(type)
+    } else {
+      seen = seen.add(type)
+    }
+    return errors
+  }, List())
+}
+
+const validateNamingCollisions = (local, foreign) =>
+  List.of(...validateNamingCollisionsInTypes(local.concat(foreign)))
```
Jannis: This could just return `validateNamingCollisionsInTypes(...)` without the `List.of(...)`

src/validation/schema.js
```diff
@@ -339,10 +582,39 @@ const validateTypeDefinition = (defs, def) =>
 const validateTypeDefinitions = defs =>
   defs.reduce((errors, def) => errors.concat(validateTypeDefinition(defs, def)), List())
 
+const validateNamingCollisionsInTypes = types => {
+  let seen = Set()
+  let errored = Set()
```
Jannis: `errored` -> `conflicting` or `knownConflicts`?

src/validation/schema.js
```diff
@@ -339,10 +582,39 @@ const validateTypeDefinition = (defs, def) =>
...
+  return types.reduce((errors, type) => {
+    if (seen.has(type) && !errored.has(type)) {
+      errors = errors.push(
+        immutable.fromJS({
+          loc: { start: 1, end: 1 },
+          entity: type,
+          message: `Type ${type} is defined more than once`,
```
Jannis: I have the feeling that this error message is not going to survive very long, as it will not be helpful enough.

It may be worth to maintain the notion of local and imported types here and tell users when a type is defined multiple times in the subgraph schema or when a type conflicts with another schema (and where that was imported from).

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
-    ),
+    def.name && def.name.value == SUBGRAPH_SCHEMA
+      ? List.of(...validateSubgraphSchemaDirectives(def), ...validateTypeHasNoFields(def))
+      : List.of(
+          ...validateEntityDirective(def),
+          ...validateEntityID(def),
+          ...validateEntityFields(defs, def),
+          ...validateNoImportsDirective(def),
+        ),
+  ObjectTypeExtension: (_defs, def) => List.of(...validateAtLeastOneExtensionField(def)),
```
Jannis: The `List.of()` is not necessary

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+    ? List([
+        immutable.fromJS({
+          loc: def.name.loc,
+          entity: def.name.value,
+          message: `${def.name.value} type is not allowed any fields by convention`,
+        }),
+      ])
+    : List()
+
+const validateAtLeastOneExtensionField = def => List()
```
Jannis: I assume this is a placeholder for the `extend type` phase of this feature implementation?

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+
+const validateImportDirectiveFrom = (def, directive) => {
+  let from = directive.arguments.find(argument => argument.name.value == 'from')
+  return from
+    ? validateImportDirectiveArgumentFromIsValid(def, directive, from)
+    : List([
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message: `@import argument 'from' required: specify the subgraph to import types from`,
```
Jannis: Nit: we're using inconsistent language here: `must be specified` vs. `required`

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+          message: `@imports directive from: only fields 'id' or 'name' allowed`,
+        }),
+      )
+    }
+    if (field.value.kind != 'StringValue') {
+      return errors.push(
+        immutable.fromJS({
+          loc: field.name.loc,
+          entity: def.name.value,
+          message: '@imports directive from: fields must have string values',
```
Jannis: Inconsistent language: `argument: 'from'` vs. `directive from:`

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+}
+
+const validateImportDirectiveType = (def, directive, type) =>
+  importDirectiveTypeValidators[type.kind]
+    ? importDirectiveTypeValidators[type.kind](def, directive, type)
+    : List().push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message: 'Imported type must be either String | {name: String, as: String}',
```
Jannis: think the syntax could be made more human-readable, e.g.
```js
'Imported type must be one of "Name", { name: "Name" } or { name: "Name", as: "Alias" }'
```

src/validation/schema.js
Jannis: In the `types` validation you mentioned which fields there are. I think it would be good to be specific here as well if we can.

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+        }),
+      )
+
+const validateImportDirectiveArgumentTypesIsValid = (def, directive, argument) => {
+  if (argument.value.kind != 'ListValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: types must be an list',
```
Jannis: Let's `types` in backticks or quotes to be consistent. It would be good to make the errors consistent across the board in the changes.

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+const importDirectiveTypeValidators = {
+  StringValue: (_def, _directive, _type) => List(),
+  ObjectValue: (def, directive, type) => {
+    let errors = List()
+    if (type.fields.length != 2) {
+      return errors.push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message: '@imports type objects accept and require two fields: [name, as]',
```
Jannis: This is confusing because `as` is not required. Let's just be specific about the forms possible (see my other comment further below).

src/validation/schema.js
```diff
@@ -322,13 +378,200 @@ const validateEntityFields = (defs, def) =>
...
+        immutable.fromJS({
+          loc: def.name.loc,
+          entity: def.name.value,
+          message: `@imports directive only allowed on '${SUBGRAPH_SCHEMA}' type`,
+        }),
+      )
+    : List()
+
+const importDirectiveTypeValidators = {
+  StringValue: (_def, _directive, _type) => List(),
```
Jannis: Is `StringValue` a thing? Isn't t just `String`?

src/validation/schema.js
Jannis: What do you think? I'd strongly prefer that over using a special `_SubgraphSchema_` type. I think we could reuse most of the validation code here even.

Jannis:
Oh yes, some tests would be good in https://github.com/graphprotocol/graph-cli/blob/master/src/validation/schema.test.js.

src/validation/schema.js
olibearo: Where is the field required? Parsing works without the field and this type will be dropped in the schema merge.

Also, what's the draw of putting it in a comment? That is more jarring to me than a type (which we can rename so that it reads closer to its intent) with only imports on it.

src/validation/schema.js
olibearo: Indeed

src/validation/schema.js
olibearo: It is indeed `StringValue`

src/validation/schema.js
```diff
@@ -38,6 +38,9 @@ const TYPE_SUGGESTIONS = [
   ],
 ]
 
+// As a convention, the type _schema_ is reserved to define imports on.
+const RESERVED_TYPE = '_schema_'
```
Jannis: We landed on `type _Schema_` in the end, with a capital `S`.

src/validation/schema.js
```diff
@@ -154,6 +157,65 @@ const unwrapType = type => {
...
+  defs
+    .filter(
+      def =>
+        def.kind === 'ObjectTypeDefinition' ||
+        def.kind === 'EnumTypeDefinition' ||
+        def.kind === 'InterfaceTypeDefinition',
+    )
+    .map(def => def.name.value)
+
+const gatherForeignTypes = defs =>
```
Jannis: I've reworked the terminology in the RFC a little, let's call these types `imported types` and this function `gatherImportedTypes` now.

src/validation/schema.js
```diff
@@ -154,6 +157,65 @@ const unwrapType = type => {
...
+  defs
+    .filter(
+      def =>
+        def.kind === 'ObjectTypeDefinition' &&
+        def.name.value == RESERVED_TYPE &&
+        def.directives.find(
+          directive =>
+            directive.name.value == 'imports' &&
```
Jannis: I think the directive should be `@import`, not `@imports`. Similar to how in other languages you'd do `import ... from ...` and not `imports ... from ...`.

src/validation/schema.js
```diff
@@ -154,6 +157,65 @@ const unwrapType = type => {
...
+            directive.arguments.find(argument => argument.name.value == 'types'),
+        ),
+    )
+    .map(def =>
+      def.directives
+        .filter(
+          directive =>
+            directive.name.value == 'imports' &&
```
Jannis: Same here

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
     List(),
   )
 
+const validateNoImportsDirective = def =>
+  def.directives.find(directive => directive.name.value == 'imports')
```
Jannis: And here

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
     List(),
   )
 
+const validateNoImportsDirective = def =>
```
Jannis: `NoImports` -> `NoImport`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+  def.directives.find(directive => directive.name.value == 'imports')
+    ? List().push(
+        immutable.fromJS({
+          loc: def.name.loc,
+          entity: def.name.value,
+          message: `@imports directive only allowed on '${RESERVED_TYPE}' type`,
```
Jannis: `@imports` -> `@import`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+  StringValue: (_def, _directive, _type) => List(),
+  ObjectValue: (def, directive, type) => {
+    let errors = List()
+    if (type.fields.length != 2) {
+      return errors.push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message:
+            'Imported type must be one of "Name", { name: "Name" }, or { name: "Name", as: "Alias" }',
```
Jannis: `Imported type` -> `Import`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+        }),
+      )
+    }
+    return type.fields.reduce((errors, field) => {
+      if (!['name', 'as'].includes(field.name.value)) {
+        return errors.push(
+          immutable.fromJS({
+            loc: directive.name.loc,
+            entity: def.name.value,
+            message: `@imports type object field '${field.name.value}' invalid,  may only be one of: [name, as]`,
```
Jannis: `@imports` -> `@import`, remove the `type object`, remove the double space after `invalid,`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+          }),
+        )
+      }
+      if (field.value.kind != 'StringValue') {
+        return errors.push(
+          immutable.fromJS({
+            loc: directive.name.loc,
+            entity: def.name.value,
+            message: '@imports type object fields [name, as] must be strings',
```
Jannis: `@imports` -> `@import`, remove the `type object`.

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+
+const validateImportDirectiveType = (def, directive, type) => {
+  return importDirectiveTypeValidators[type.kind]
+    ? importDirectiveTypeValidators[type.kind](def, directive, type)
+    : List().push(
+        immutable.fromJS({
+          loc: directive.name.loc,
+          entity: def.name.value,
+          message:
+            'Imported type must be one of "Name", { name: "Name" }, or { name: "Name", as: "Alias" }',
```
Jannis: `Imported type` -> `Import`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+        }),
+      )
+}
+
+const validateImportDirectiveArgumentTypesIsValid = (def, directive, argument) => {
```
Jannis: I think we can drop `IsValid` here, we already say `validate` in the name.

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+  if (argument.value.kind != 'ListValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: `types` must be an list',
```
Jannis: `@imports directive argument:` -> `@import argument`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+    )
+  }
+
+  return argument.value.values.reduce(
+    (errors, type) => errors.concat(validateImportDirectiveType(def, directive, type)),
+    List(),
+  )
+}
+
+const validateImportDirectiveArgumentFromIsValid = (def, directive, argument) => {
```
Jannis: Drop `IsValid`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+  if (argument.value.kind != 'ObjectValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: '@imports directive argument: `from` must be an object',
```
Jannis: `@imports directive argument:` -> `@import argument`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+      }),
+    )
+  }
+
+  if (argument.value.fields.length != 1) {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: `@imports directive argument: 'from' must have one field of id or name`,
```
Jannis: `@imports directive argument:` -> `@import argument`

`field of id or name` -> improve to ```must have an `id` or `name` field```

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+    )
+  }
+
+  return argument.value.fields.reduce((errors, field) => {
+    if (!['name', 'id'].includes(field.name.value)) {
+      return errors.push(
+        immutable.fromJS({
+          loc: field.name.loc,
+          entity: def.name.value,
+          message: `@imports directive argument: 'from' must be one of { name: "Name" } or { id: "ID" }`,
```
Jannis: `@imports directive argument:` -> `@import argument`

src/validation/schema.js
Jannis: Also: `@imports directive from:` -> `@import argument 'from': ....`?

src/validation/schema.js
Jannis: Simplify to `@import argument 'from' must be specified`, done.

src/validation/schema.js
Jannis: Ah, of course it is. Sorry. 😄

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+  )
+}
+
+const validateImportDirectiveArgumentFrom = (def, directive, argument) => {
+  if (argument.value.kind != 'ObjectValue') {
+    return List().push(
+      immutable.fromJS({
+        loc: directive.name.loc,
+        entity: def.name.value,
+        message: `@import directive argument: 'from' must be an object`,
```
Jannis: drop `directive`

src/validation/schema.js
```diff
@@ -322,13 +378,203 @@ const validateEntityFields = (defs, def) =>
...
+          message: `@import argument: 'from' must be one of { name: "Name" } or { id: "ID" }`,
+        }),
+      )
+    }
+    if (field.value.kind != 'StringValue') {
+      return errors.push(
+        immutable.fromJS({
+          loc: field.name.loc,
+          entity: def.name.value,
+          message: `@import argument: 'from' must be one of { name: "Name" } or { id: "ID" } with string values`,
```
Jannis: I'd drop the `:` after all these messages. Like below where we have `@import argument 'types' must be specified`, I think these can all be `@import argument 'from' must be one of...`. The read flow is better that way, IMHO.

PR #398, opened 2019-11-21
Fix contract call codegen for small uints

### LINKED ISSUE #92: [Types][uint8] Bad AssemblyScript type generation for uint8 smart contract type
Hi everyone!
I have a problem when I try to use an smart contract function which an **uint8 type as an input parameter**.

When I use that function from one of my mappings after deploying it to thegraph.com website it throws this error:

`Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleAMMOutcomeTokenTrade": Failed to call function "calcMarginalPrice" of contract "LMSRMarketMaker": type mismatch, token Int(0) is not of kind Uint(8), code: SubgraphSyncingFailure, id: QmSMt6hZjnTV4ehVawkpimNPXqA6yLywYfCNfoaGxUkXyd`

Let me explain my specific case:

1. **ABI**. Function definition which the **input parameter** that I am talking about:
```
  {
    "constant": true,
    "inputs": [{ "name": "outcomeTokenIndex", "type": "uint8" }],
    "name": "calcMarginalPrice",
    "outputs": [{ "name": "price", "type": "uint256" }],
    "payable": false,
    "stateMutability": "view",
    "type": "function"
  }
```

2. **AssemblyScript code generation**. When I execute `yarn codegen` it generates this result:
```
calcMarginalPrice(outcomeTokenIndex: i32): BigInt {
    let result = super.call("calcMarginalPrice", [
      EthereumValue.fromI32(outcomeTokenIndex)
    ]);
    return result[0].toBigInt();
  }
```

3. **Function call**. This is how I call the previous generated function:
```
  let outcomeSlotCountI32 = entity.outcomeSlotCount.toI32()
  let marketMakerMarginalPrices = new Array<BigInt>(outcomeSlotCountI32);
  for (let outcomeSlotIndex = 0; outcomeSlotIndex < outcomeSlotCountI32; ++outcomeSlotIndex) {
    marketMakerMarginalPrices[outcomeSlotIndex] = lmsrMarketMakercontract.calcMarginalPrice(outcomeSlotIndex);
  }
```

4.  **thegraph.com error**. This the error that I can see when I deploy my subgraph:
`Subgraph instance failed to run: Failed to process trigger: Failed to handle Ethereum event with handler "handleAMMOutcomeTokenTrade": Failed to call function "calcMarginalPrice" of contract "LMSRMarketMaker": type mismatch, token Int(0) is not of kind Uint(8), code: SubgraphSyncingFailure, id: QmSMt6hZjnTV4ehVawkpimNPXqA6yLywYfCNfoaGxUkXyd`

5. **Temporal solution**. While I was wondering about that error, I realized that it is fixed if I change the **fromI32** function definition in the `graph-ts/index.ts file` like this (**token.kind value** was replaced from EthereumValueKind.INT to EthereumValueKind.**UINT**):
```
  static fromI32(i: i32): EthereumValue {
    let token = new EthereumValue()
    token.kind = EthereumValueKind.UINT
    token.data = BigInt.fromI32(i) as u64
    return token
  }
```

What do you think? Could it be a bug in the AssemblyScript code generation or shall I call the function in another way?

Thank you!

leoyvens:
We need the `EthereumValueKind` to be `UINT` so that the call decodes correctly, this can be achieved with a small change to codegen.

Resolves https://github.com/graphprotocol/graph-ts/issues/92.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,15 +1,18 @@
-dist: trusty
+# Note: Commented out things were to support the `graph auth` test, which
+# stopped working after upgrading the dist to Xenial (or Bionic).
+
+dist: bionic
 language: node_js
 node_js:
-  - "10"
-  - "11"
+  - '10'
+  - '11'
 cache: yarn
-before_install:
-  - sudo apt-get install libsecret-1-dev dbus gnome-keyring python-keyring python-gnomekeyring
-before_script:
+# before_install:
+#  - sudo apt-get install libsecret-1-dev dbus gnome-keyring python-keyring python-gnomekeyring
+# before_script:
   # Create a new 'login' keyring; this appearas to be necessary for gnome-keyring to auto-launch
   # properly when needed
-  - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
+  # - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
 script:
   - yarn test
   - cd examples/example-subgraph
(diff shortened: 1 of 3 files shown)


Jannis (approved):
Looks good to me!

leoyvens:
Also fixed Travis by updating the Linux dist and disabling the graph auth test.

PR #406, opened 2019-12-13
Mutations Support

dOrgJelli:
With these changes the `graph build` command now supports subgraph mutations.

### Here's an example project setup:
**subgraph.yaml**
```yaml
...
schema:
  file: ./schema.graphql
mutations:
  file: ./src/mutations/mutations.yaml
dataSources:
  ...
```
**mutations.yaml**
```yaml
specVersion: 0.0.1
repository: https://npmjs.com/package/gravatar-mutations
schema:
  file: ./schema.graphql
resolvers:
  kind: javascript
  file: ./dist/index.js
```

**src/mutations/schema.graphql**
```graphql
input Options {
  value: String!
}

type Mutation {
  someFunction(
    options: Options!
  ): EntityType
}
```

**schema.graphql**
```graphql
type EntityType @entity {
  id: ID!
}
```

### The following syntax is also supported:
**subgraph.yaml**
```yaml
...
schema:
  file: ./schema.graphql
mutations:
  specVersion: 0.0.1
  repository: https://npmjs.com/package/gravatar-mutations
  schema:
    file: ./src/mutations/schema.graphql
  resolvers:
    kind: javascript
    file: ./src/mutations/dist/index.js
dataSources:
  ...
```


manifest-schema.graphql
```diff
@@ -6,6 +6,10 @@ type Schema {
   file: File!
 }
 
+type Manifest {
+  file: File!
+}
+
```
Jannis: I'd call this `ExternalMutations` and the inline `MutationsManifest` -> `InlineMutations`.

manifest-schema.graphql
```diff
@@ -80,3 +85,18 @@ type EthereumContractDataSourceTemplate {
 type EthereumContractSourceTemplate {
   abi: String!
 }
+
+union Mutations = Manifest | MutationsManifest
+union Resolvers = JavascriptResolvers
```
Jannis: Let's drop this union type, I don't think we need it (yet).

manifest-schema.graphql
```diff
@@ -80,3 +85,18 @@ type EthereumContractDataSourceTemplate {
...
+
+type JavascriptResolvers {
+  kind: String!
+  file: File!
+}
```
Jannis: Let's call this one `MutationResolvers`

Jannis: Or `MutationsResolvers` (harder on the tongue). I'm ok with either.

src/compiler.js
```diff
@@ -104,6 +104,21 @@ class Compiler {
...
+        let mutations = subgraph.get('mutations')
+        files.push(mutations.getIn(['schema', 'file']))
+
+        let resolversKind = mutations.getIn(['resolvers', 'kind'])
+        switch (resolversKind) {
+          case "javascript":
+            files.push(mutations.getIn(['resolvers', 'file']))
+            break;
+          default:
+            throw Error(`Unimplemented resolvers kind '${resolversKind}'`)
```
Jannis: I think it's safe to assume that all resolver types would be defined in an external file. So we can drop the `switch` here.

We should catch unsupported resolver types when we validate the manifest.

src/compiler.js
```diff
@@ -480,6 +495,64 @@ class Compiler {
...
+              // Write the mutation schema + root schema to the output directory
+              .updateIn(['schema', 'file'], schemaFile => {
+                const rootSchemaFile = path.resolve(
+                  this.sourceDir, subgraph.getIn(['schema', 'file'])
+                )
+                schemaFile = path.resolve(this.sourceDir, schemaFile)
+
+                // Concatenate the schemas
+                let schemaData = fs.readFileSync(rootSchemaFile, 'utf-8')
+                schemaData += '\n' + fs.readFileSync(schemaFile, 'utf-8')
```
Jannis: Might as well do this in one statement:
```js
let schema = fs.readFileSync(rootSchemaFile, 'utf-8') + '\n' + fs.readFileSync(schemaFile, 'utf-8')
```

Jannis: (I'm coming from function languages, where mutating variables are frowned upon; that's why I prefer not touching variables after they have been defined.)

src/compiler.js
```diff
@@ -480,6 +495,64 @@ class Compiler {
...
+
+                // Write the result to build/mutations/schema.graphql
+                return path.relative(
+                  this.options.outputDir,
+                  this._writeSubgraphFile(
+                    'mutations/schema.graphql',
```
Jannis: Not going to work on Windows, we need to do `path.join('mutations', 'schema.graphql')` here.

src/compiler.js
```diff
@@ -480,6 +495,64 @@ class Compiler {
...
+              // Write the resolvers file
+              .updateIn(['resolvers'], resolvers => {
+                const resolversKind = resolvers.get('kind')
+                switch (resolversKind) {
+                  case 'javascript':
+                    return resolvers.update('file', file => {
+                      const resolversData = fs.readFileSync(
+                        path.resolve(this.sourceDir, file), 'utf-8'
+                      )
+
```
Jannis: Drop the `switch` (see my earlier comment about it).

src/compiler.js
```diff
@@ -480,6 +495,64 @@ class Compiler {
...
+                      return path.relative(
+                        this.options.outputDir,
+                        this._writeSubgraphFile(
+                          'mutations/resolvers/index.js',
```
Jannis: Use `path.join()` instead of hard-coding `/` in this path.

src/compiler.js
```diff
@@ -480,6 +495,64 @@ class Compiler {
               )
         })
 
+        if (subgraph.has('mutations')) {
+          // Copy mutation files and update their paths
+          subgraph = subgraph.update('mutations', mutations =>
+            // mutations/schema.graphql + mutations/resolvers/index.js
```
Jannis: Why not just `mutations/resolvers.js` or `mutations/index.js` instead of adding the extra `resolvers` directory?

src/compiler.js
```diff
@@ -561,6 +634,25 @@ class Compiler {
           })
         }
 
+        // Upload mutations schema and resolvers if present
+        if(subgraph.get('mutations')){
```
Jannis: The `if(` and `)){` suggest this file wasn't formatted with Prettier. 😉

src/subgraph.js
```diff
@@ -111,6 +117,25 @@ module.exports = class Subgraph {
...
+
+      if (errorMsg) {
+        return immutable.fromJS([
+          {
+            path: ['mutations', 'resolvers', 'file'],
+            message: errorMsg
+          }
+        ])
+      }
+    }
```
Jannis: We try to collect as many errors as possible at once to reduce the amount of trial-and-error users have to go through. If `validateMutationResolvers` can return more than one error, we should have it return a list of errors and not just an error message.

src/validation/manifest.js
```diff
@@ -51,13 +51,112 @@ const validators = immutable.fromJS({
...
+              path: [...ctx.get('path'), 'kind'],
+              message: `No value provided`,
+            },
+          ])
+        } else if (value.get('kind') === "javascript") {
+          unionType = "JavascriptResolvers"
+        } else {
+          return immutable.fromJS([
+            {
+              path: [...ctx.get('path'), 'kind'],
```
Jannis: I can see how we'd want a more specific path when validating union type values, but I disagree with how it's done here. This `UnionTypeDefinition` dispatcher is supposed to be generic. Data sources and templates were already being validated before, there is no need to hard-code anything about them here, I think.

We should instead have two specific additional type validators: One for `ExternalMutations` and one for `InlineMutations`. (Ok, and one for the resolvers.)

src/validation/resolvers.js
```diff
@@ -0,0 +1,95 @@
+const fs = require('fs')
```
Jannis: I'm going to postpone reviewing this file until we have the spec as an RFC in https://github.com/graphprotocol/rfcs/.

src/validation/schema.js
```diff
@@ -339,9 +380,63 @@ const validateTypeDefinition = (defs, def) =>
...
+  List.of(
+    ...validateNoOtherTypesExceptMutation(defs),
+    ...validateMutationTypeExists(defs)
+  )
+
+const validateSchema = (rootFile, mutationsFile) => {
+  let schemas = [rootFile]
+
+  // If the mutation schema is provided, concatenate it
+  // to the end of the root schema, and it separately
```
Jannis: `and it separately`? 😁

src/validation/schema.js
```diff
@@ -47,9 +47,14 @@ const typeSuggestion = typeName =>
...
-const loadSchema = filename => {
+const loadSchema = filenames => {
   try {
-    return fs.readFileSync(filename, 'utf-8')
+    // Concatenate multiple schema files together
+    let result = ""
+    for (filename of filenames) {
+      result += fs.readFileSync(filename, 'utf-8')
+    }
+    return result;
```
Jannis: Being the functional programmer that I am, I prefer this:
```js
return filenames
  .map(filename => fs.readFileSync(filename, 'utf-8'))
  .join('\n')
```

Jannis: But I'd rather stick to loading a single schema in `loadSchema`, instead of loading many files at once.

src/validation/schema.js
```diff
@@ -339,9 +380,63 @@ const validateTypeDefinition = (defs, def) =>
...
+    ...validateMutationTypeExists(defs)
+  )
+
+const validateSchema = (rootFile, mutationsFile) => {
+  let schemas = [rootFile]
+
+  // If the mutation schema is provided, concatenate it
+  // to the end of the root schema, and it separately
+  if (mutationsFile) {
+    schemas.push(mutationsFile)
```
Jannis: Drop this, instead make `schemas` an array of already loaded schemas.

src/validation/schema.js
```diff
@@ -339,9 +380,63 @@ const validateTypeDefinition = (defs, def) =>
...
+        ]))
+      : errors,
+    List()
+  )
+
+const validateMutationTypeExists = defs =>
+  defs.findIndex(def => def.name.value === "Mutation") === -1
+    ? immutable.fromJS([
+        {
+          loc: "",
```
Jannis: I _think_ (not sure), `loc: undefined` or `loc: null` are better here.

src/validation/schema.js
```diff
@@ -339,9 +380,63 @@ const validateTypeDefinition = (defs, def) =>
...
+    def.kind === "ObjectTypeDefinition" &&
+    def.name.value !== "Mutation"
+      ? errors.concat(immutable.fromJS([
+          {
+            loc: def.loc,
+            entity: def.name.value,
+            message:
+              `ObjectTypeDefinition '${def.name.value}' found, Mutation ` +
+              "must be the only type. Try using 'input' for mutation arguments, " +
+              "'interface' for mutation return values, or existing @entity types from the subgraph schema.",
```
Jannis: Let's unravel this a little bit. How about this error mesage?

```js
`\
Encountered object type '${def.name.value}. 'Mutation' is the only \
object type allowed in the mutations schema.

For mutation arguments, use 'input' types. For return values, use either \
entity types from the subgraph schema or interfaces.`
```

I actually think that we should allow other, non-entity types as return values here, and should allow users to define these types in the mutations schema.

The combined schema validation would have to become a little smarter to handle these. One trick could be to transform the type definitions in the mutations schema to add a `@mutations` directive to all of them. In the combined schema validation, we could then ignore all types with this directive.

tests/cli/validation.test.js
```diff
@@ -9,6 +9,206 @@ describe('Validation', () => {
       exitCode: 1,
     },
   )
+  cliTest(
```
Jannis: I'm going to skip reviewing the tests for now, because they may change a little after the code review has been addressed. I'm glad to see so many tests though!

Jannis:
One more comment: We have an (undocumented) convention for our commit messages. The commit subject is expected to start with the "module" or scope that the changes are made against. Would it be possible to follow this convention in this PR as well?

Check some of the more recent commit messages for reference:

```
travis: Upgrade the Linux dist and disable the `graph auth` test
codegen: Fix contract call codegen for small uints
compiler: Only compile identical mappings once
compiler: Cache uploaded files using SHA1 instead of paths (#393)
commands/init: Add POA Core support
scaffold: Bump graph-ts to 0.16.0
NPM: Add release:patch and release:minor commands
examples/example-subgraph: Bump graph-cli and graph-ts
tests/cli/validation: Add tests for network validation
commands: Fix spacing
commands: Use the correct exit codes on deploy cmd
manifest-schema, manifest: Add startBlock field, validation for BigInt
compiler: Improve error message for not found `node_modules`
compiler: Improve loop to find `node_modules`
compiler: Increase number of ancestor directories checked for libs
compiler: Check directories above for libs
compiler: Prefer `print.error` over `console.log`
compiler: Do not assume the source dir is the working dir
examples, scaffold: Update to graph-ts 0.15.0
example-subgraph: Update graph-ts
```

src/compiler.js
dOrgJelli: I was thinking that the extra directory would be nice if we ended up having additional files alongside the index.js. The main example I was thinking of was type definitions, so index.d.ts. I suppose these both could be included at the root directory and it would still be fine, as name conflicts shouldn't be an issue. I'll move it up to the root directory.

src/validation/manifest.js
dOrgJelli: The reason we went about writing this is because the original UnionTypeDefinition wasn't actually functional for Unions of multiple types (code [here](https://github.com/graphprotocol/graph-cli/blob/a4b62fab8138e298455e96ad0acba95a5eecfd4a/src/validation/manifest.js#L56)). It would iterate through all union types, and aggregate the errors from each one. Instead of doing this, we instead want to have some way of determine what union type we need to use. This is done by looking at the `kind` property in some cases (datasource types for example), and also by looking for the presence of a special determining field like the `file` field in the `ExternalMutations` type.

It is for the above reasons that I think we need to have programmatic type deducers. Would you agree? Another route is to iterate through all union types, and if one succeeds, use that as the resulting type and disregard all other errors. This would lead to some weird errors being returned if none succeed, as all possible errors form all possible types would be returned.

src/validation/schema.js
dOrgJelli: Re @mutations directive: This would require changes within the graph-node to ignore these new types correct?

dOrgJelli:
@Jannis in order to follow the commit message standard, I can recreate this PR from a new branch where I commit the file changes one by one. Does that sound like a good choice to you?

tests/cli/validation.test.js
dOrgJelli: Tests have been updated.

src/validation/schema.js
dOrgJelli: The error message has been updated with the version you wrote, thank you for simplifying the wording!

Jannis:
@dOrgJelli Why not rebase and reword all commit messages as you go?

dOrgJelli:
@Jannis commit messages have been updated

azf20:
Closing as stale, but definitely interested in reigniting mutation discussions!

PR #414, opened 2020-01-06
Add `graph test` command

Jannis:
This adds a new `graph test` command that can be used to run test commands against a (customizable) test environment (by default: Graph Node, Postgres, IPFS and Ganache in Docker Compose).

This provides a rough framework for writing integration tests where a subgraph is indexed against a fresh Graph node and tests can wait for it to be synced, run queries at specific blocks etc. It is a bit rough around the edges, so I'm tempted to move it into a `beta` subcommand, to be invoked as `graph beta test [options] <test-command>`.

# Usage

```
graph test [options] <test-command>

Options:

  -h, --help                    Show usage information
      --compose-file <file>     Custom Docker Compose file for additional services (optional)
      --node-image <image>      Custom Graph Node image to test against (default: graphprotocol/graph-node:latest)
      --standalone-node <cmd>   Use a standalone Graph Node outside Docker Compose (optional)
      --standalone-node-args    Custom arguments to be passed to the standalone Graph Node (optional)
      --skip-wait-for-ipfs      Don't wait for IPFS to be up at localhost:5001 (optional)
      --skip-wait-for-ethereum  Don't wait for Ethereum to be up at localhost:8545 (optional)
      --skip-wait-for-postgres  Don't wait for Postgres to be up at localhost:5432 (optional)
      --node-logs               Print the Graph Node logs (optional)
```

The `<test-command>` could be anything: it could be a shell script that builds and deploys a subgraph, it could be a JS test suite run with `jest` or it could be a Truffle test suite, run with `truffle test`.

Overriding the `--node-image` allows to use a custom build of Graph Node, which would be useful for running integration tests for every commit of Graph Node itself.

Overriding the `--compose-file` allows to e.g. use a different Ethereum provider than the default Ganache, making it possible to test subgraphs against `mainnet` for instance.

## Example usage

Inside subgraph repository with contracts, Truffle migrations and Truffle tests:

```sh
graph test "truffle test"
```

Such an example setup can be found in https://github.com/graphprotocol/graph-cli/blob/jannis/graph-test/examples/basic-event-handlers/.

* Subgraph: https://github.com/graphprotocol/graph-cli/blob/jannis/graph-test/examples/basic-event-handlers/subgraph.yaml
* Truffle migrations: https://github.com/graphprotocol/graph-cli/tree/jannis/graph-test/examples/basic-event-handlers/migrations
* Truffle config: https://github.com/graphprotocol/graph-cli/blob/jannis/graph-test/examples/basic-event-handlers/truffle.js
* Tests: https://github.com/graphprotocol/graph-cli/blob/jannis/graph-test/examples/basic-event-handlers/test/handlers.js. 

What `graph test "truffle test --network test"` in this example does is:

1. Start up the test environment (Graph Node, Ganache etc.)
2. Deploy the contracts to Ganache
3. Patch the subgraph manifest with the right addresses
4. Deploy the subgraph
5. Wait for it to be indexed
6. Run queries and verify their results
7. Tear down the test environment again

Steps 1 and 7 are provided by `graph test`. Step 2 is handled by Truffle. Steps 3-6 have to be implemented by the users themselves; we could provide more tooling for this over time.

## Testing `graph test` itself

The above `basic-event-handlers` example is included in the Travis pipeline of `graph-cli`. The output from one of the previous test runs can be seen here: https://travis-ci.org/graphprotocol/graph-cli/jobs/633258441#L536-L568


Jannis:
I've added the following CLI flags:

`--standalone-node` to run tests using a Graph node on the host
`--standalone-node-args` to replace the default arguments passed to the standalone Graph node
`--skip-wait-for-ipfs` to skip waiting for IPFS to become available on localhost:5001
`--skip-wait-for-ethereum` to skip waiting for Ethereum to become available on localhost:8545
`--skip-wait-for-postgres` to skip waiting for Postgres to become available on localhost:5432

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+const net = require('net')
+const tmp = require('tmp')
+const Docker = require('dockerode')
+const path = require('path')
+const stripAnsi = require('strip-ansi')
+const util = require('util')
+const exec = util.promisify(require('child_process').exec)
+const spawn = require('child_process').spawn
+
+const { createCompiler } = require('../command-helpers/compiler')
```
davekaj: Unused variable

examples/basic-event-handlers/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "devDependencies": {
+    "@graphprotocol/graph-cli": "^0.17.1",
+    "@graphprotocol/graph-ts": "^0.17.0",
+    "apollo-fetch": "^0.7.0"
+  },
+  "dependencies": {
+    "babel-polyfill": "^6.26.0",
+    "babel-register": "^6.26.0",
+    "truffle": "^5.0.4",
+    "truffle-contract": "^4.0.5",
```
davekaj: I don't believe this package is used anywhere

examples/basic-event-handlers/contracts/Gravity.sol
```diff
@@ -0,0 +1,62 @@
+pragma solidity ^0.4.0;
```
davekaj: This is pretty old solidity. Same with the truffle packages used. 

I don't believe it really matters. AFAIK the solidity compiler is backwards compatible and shouldn't effect our testing at all. But I figured I would point it out, in case you think it could effect us

examples/basic-event-handlers/test/handlers.js
```diff
@@ -0,0 +1,110 @@
...
+const GravatarRegistry = artifacts.require('./GravatarRegistry.sol')
+
+const srcDir = path.join(__dirname, '..')
+
+const fetchSubgraphs = createApolloFetch({ uri: 'http://localhost:8030/graphql' })
+const fetchSubgraph = createApolloFetch({
+  uri: 'http://localhost:8000/subgraphs/name/test/basic-event-handlers',
+})
+
+const waitForSubgraphToBeSynced = async () =>
```
davekaj: This function looks good. But I wonder, what if someone is trying to sync a subgraph that is going to take hours or days? and they want to test it.

I'm thinking two things. One - we allow them to point to an already fully synced subgraph.

Or if we want to test specific handlers on early blocks , we could designate a block, like 8,000,000, and then just test the first time each handler runs. 

I guess this is only relevant if we are allowing testing on mainnet. Since ganache will be very easy to sync

examples/basic-event-handlers/test/handlers.js
```diff
@@ -0,0 +1,110 @@
...
+contract('Basic event handlers', accounts => {
+  // Deploy the subgraph once before all tests
+  before(async () => {
+    // Deploy the contract
+    const registry = await GravatarRegistry.deployed()
+
+    // Insert its address into subgraph manifest
+    await patching.replace(
+      path.join(srcDir, 'subgraph.yaml'),
+      '0x2E645469f354BB4F5c8a05B3b30A929361cf77eC',
```
davekaj: If we want, we can deterministically determine the addresses. Instead of requiring the user to manually do it. See https://ethereum.stackexchange.com/questions/760/how-is-the-address-of-an-ethereum-contract-computed

As long as we can get the deploying account (which we can from truffle), we can do hashes of the account and its nonce, and deterministically fill in the addresses

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+      --skip-wait-for-ipfs      Don't wait for IPFS to be up at localhost:5001 (optional)
+      --skip-wait-for-ethereum  Don't wait for Ethereum to be up at localhost:8545 (optional)
+      --skip-wait-for-postgres  Don't wait for Postgres to be up at localhost:5432 (optional)
+`
+
+module.exports = {
+  description: 'Tests the current project against a Graph Node and Parity testnet',
+  run: async toolbox => {
+    // Obtain tools
+    let { filesystem, print, system } = toolbox
```
davekaj: system is an unused variable

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+      --node-image <image>      Custom Graph Node image to test against (default: graphprotocol/graph-node:latest)
+      --standalone-node <cmd>   Use a standalone Graph Node outside Docker Compose (optional)
+      --standalone-node-args    Custom arguments to be passed to the standalone Graph Node (optional)
+      --skip-wait-for-ipfs      Don't wait for IPFS to be up at localhost:5001 (optional)
+      --skip-wait-for-ethereum  Don't wait for Ethereum to be up at localhost:8545 (optional)
+      --skip-wait-for-postgres  Don't wait for Postgres to be up at localhost:5432 (optional)
+`
+
+module.exports = {
+  description: 'Tests the current project against a Graph Node and Parity testnet',
```
davekaj: Do you mean ganache testnet?

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+      try {
+        nodeProcess = await startGraphNode(
+          toolbox,
+          standaloneNode,
+          standaloneNodeArgs,
+          nodeOutputChunks,
+        )
+      } catch {
+        toolbox.print.error('')
+        toolbox.print.error('  Graph Node')
```
davekaj: I think we need better error messages here. I just got:

```
  Graph Node
  ----------

  Attaching to
```
And an exit. There are three instances of `Graph Node` with the line under it being printed, so it is also unclear which place this error is coming from (although in my case it  the fact that Graph Node could not be queried)

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+
+    if (result.exitCode !== 0 || nodeLogs) {
+      toolbox.print.info('')
+      toolbox.print.info('  Graph Node')
+      toolbox.print.info('  ----------')
+      toolbox.print.info('')
+      toolbox.print.info(
+        indent(
+          '  ',
+          await graphNodeLogs(toolbox, tempdir, standaloneNode, nodeOutputChunks),
```
davekaj: This is actually where mine failed. I believe what happened was because there was no graph no logs since graph node didn't start up, it didn't give any message and it errored out

davekaj: Here is the whole output:
```
✖ Tests failed

  Output
  ------


  Compiling your contracts...
  ===========================
  > Compiling ./contracts/Gravity.sol
  > Compiling ./contracts/Migrations.sol



    Contract: Basic event handlers
      1) "before all" hook


    0 passing (6s)
    1 failing

    1) Contract: Basic event handlers
         "before all" hook:
       Uncaught FetchError: request to http://localhost:8030/graphql failed, reason: socket hang up
        at ClientRequest.<anonymous> (node_modules/cross-fetch/node_modules/node-fetch/index.js:133:11)
        at Socket.socketOnEnd (_http_client.js:440:9)
        at endReadableNT (_stream_readable.js:1183:12)
        at processTicksAndRejections (internal/process/task_queues.js:80:21)





  Errors
  ------



  Graph Node
  ----------
```

I think there could be another line added that supplements `Output`, `Errors`, `Graph Node` , but isn't dependant on results. But I will leave it up to you to decide

src/commands/test.js
```diff
@@ -0,0 +1,498 @@
...
+    `Stop test environment`,
+    `Failed to stop test environment`,
+    `Warnings stopping test environment`,
+    async spinner => {
+      await compose.down({ cwd: path.join(tempdir, 'compose') })
+    },
+  )
+
+const startGraphNode = async (
+  toolbox,
```
davekaj: Unused variable

src/commands/test.js
Jannis: I initially tried Parity, but I should update the description anyway.

davekay100 (changes requested):
Done the first review. 

I would also add that the README of graph-cli should include `graph test`.

It also wasn't clear to me how to run `graph test`, with using `../../bin/graph test "truffle test --network test"`

However, I think once graph-cli actually has it in master, the cli help function should suffice. However, we will need our docs updated (I can review that PR when needed)

src/commands/test.js
Jannis: Removed

examples/basic-event-handlers/package.json
Jannis: Removed

examples/basic-event-handlers/contracts/Gravity.sol
Jannis: Doesn't matter much as long as these contracts compile and deploy fine to Ganache. It's just an example subgraph for testing.

examples/basic-event-handlers/test/handlers.js
Jannis: Yeah. Again, this just an example subgraph with a basic test.

We may want to provide more convenient ways to wait for subgraphs to reach a certain block or until they sync completely, to query subgraphs etc. over time. This could be in a `@graphprotocol/graph-test` companion library or similar. For now I'm fine if all this is a bit raw.

If people wanted to test a subgraph syncing from a specific block, all they'd have to do as part of the test setup is to define a manifest with a start block.

examples/basic-event-handlers/test/handlers.js
Jannis: It's already deterministic. The Truffle setup in this example makes it so that the contract address always is `0x2E64...`.

src/commands/test.js
Jannis: Removed

src/commands/test.js
Jannis: Reworded

src/commands/test.js
Jannis: I think the problem here was that docker compose was shut down before the output was collected. This should no longer be the case, I changed how this is done in the command.

src/commands/test.js
Jannis: Let's retry this based on the latest version of this branch. It should give us more useful logs now.

src/commands/test.js
Jannis: Removed

Jannis:
@davekaj All updated.

examples/basic-event-handlers/test/handlers.js
davekaj: Hmm, yes for this example. But I was under the impression that when a user runs their own tests they would have to add in their addresses on their own. 

But now that I think about it, most the time if they are deploying a subgraph to mainnet, the addresses of the contracts already exist, so they should be hardcoded. So this is fine

Jannis: I don't want to mandate how people run their tests. Whether they use `truffle test` or something else (could even be a shell script) is is completely open. Hence this is really just an example and not a general template (yet).

davekay100 (approved):
Approving since it works on travis

PR #417, opened 2020-01-14
codegen: Fix `get` for nullable fields

leoyvens:
We forgot to check for explicit null values. This started coming up after https://github.com/graphprotocol/graph-node/pull/1428 I believe. I checked that this fixes the livepeer subgraph.


Jannis (approved):
Yep, this makes sense.

PR #443, opened 2020-03-05
Fix for subgraph.yaml not in working dir

leoyvens:
The path returned for already-built mappings shouldn't be made relative as it is already relative to the original manifest. To make that path relative to the output manifest, we must first resolve it from the source directory (where the original manifest is).

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -230,7 +230,7 @@ class Compiler {
           'Compile data source:',
           `${dataSourceName} => ${this.displayPath(outFile)} (already compiled)`,
         )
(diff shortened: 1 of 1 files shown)


src/compiler.js
```diff
@@ -230,7 +230,7 @@ class Compiler {
           'Compile data source:',
           `${dataSourceName} => ${this.displayPath(outFile)} (already compiled)`,
         )
-        return path.relative(baseDir, outFile)
+        return outFile
```
Jannis: Oh yes, this makes sense.

PR #446, opened 2020-03-11
Codegen for creating a data source with context

leoyvens:
Part of https://github.com/graphprotocol/graph-node/issues/1404.

### DIFF
--- a/src/codegen/template.js
+++ b/src/codegen/template.js
@@ -10,7 +10,7 @@ module.exports = class DataSourceTemplateCodeGenerator {
   generateModuleImports() {
     return [
       tsCodegen.moduleImports(
-        ['Address', 'DataSourceTemplate'],
+        ['Address', 'DataSourceTemplate', 'Entity'],
         '@graphprotocol/graph-ts',
       ),
     ]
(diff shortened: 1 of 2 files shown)


src/codegen/template.js
```diff
@@ -10,7 +10,7 @@ module.exports = class DataSourceTemplateCodeGenerator {
   generateModuleImports() {
     return [
       tsCodegen.moduleImports(
-        ['Address', 'DataSourceTemplate'],
+        ['Address', 'DataSourceTemplate', 'Entity'],
```
Jannis: Let's import `TypedMap` instead (see my graph-ts review).

src/codegen/template.js
```diff
@@ -49,4 +50,29 @@ module.exports = class DataSourceTemplateCodeGenerator {
...
+    let name = this.template.get('name')
+    let kind = this.template.get('kind')
+
+    switch (kind) {
+      case 'ethereum/contract':
+        return tsCodegen.staticMethod(
+          'createWithContext',
+          [
+            tsCodegen.param('address', tsCodegen.namedType('Address')),
+            tsCodegen.param('context', tsCodegen.namedType('Entity')),
```
Jannis: TypedMap? 😉

src/commands/test.js
```diff
@@ -113,6 +113,7 @@ module.exports = {
     try {
       await startTestEnvironment(tempdir)
     } catch(e) {
+      console.log(e)
```
Jannis: This would be a ```print.error(`${e}`)``` for consistency.

PR #456, opened 2020-03-30
Fix ABI generator param types for tuple arrays

### LINKED ISSUE #455: Generated function result for tuple is of undefined type
Hi, I'm having a problem with the ABI code generator where a tuple type doesn't seem to be processed correctly, and it leads to AssemblyScript errors.

See this gist: https://gist.github.com/JamesLefrere/40a7b4f6bfa0f050548faaebc1ee689f

I believe the problem occurs here in the generator: https://github.com/graphprotocol/graph-cli/blob/fa6707600be19f378ddff60757f79300f9ae86f8/src/codegen/abi.js#L247

The third argument should be the type, but it wasn't provided. 

I'm not entirely sure what the expected behaviour should be here... any ideas?

Cheers

JamesLefrere:
Locally, this seems to fix the problem described in #455. I also added an `isTupleType` util to match `isTupleArrayType`.

Before:

```typescript
export class BasketManager__getBasketResultBStruct extends ethereum.Tuple {
  get bassets(): Array<ethereum.Tuple> {
    return this[0].toTupleArray<undefined>();
```

After:

```typescript
export class BasketManager__getBasketResultBStruct extends ethereum.Tuple {
  get bassets(): Array<BasketManager__getBasketResultBBassetsStruct> {
    return this[0].toTupleArray<BasketManager__getBasketResultBBassetsStruct>(); // This struct was generated, too
```

Before:

```typescript
export class BasketManager__getBassetsResult {
  value0: Array<ethereum.Tuple>;
```

After:

```typescript
export class BasketManager__getBassetsResult {
  value0: Array<BasketManager__getBassetsResultBAssetsStruct>;
```

Resolves #455

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -229,7 +229,7 @@ module.exports = class AbiCodeGenerator {
     }
 
     // Generate getters and classes for the param (classes only created for Ethereum tuple types)
-    return valueType === 'tuple'
+    return util.containsTupleType(valueType)
       ? this._generateTupleType(
           inputOrOutput,
           index,
@@ -245,7 +245,7 @@ module.exports = class AbiCodeGenerator {
             typesCodegen.ascTypeForEthereum(valueType),
             `
             return ${typesCodegen.ethereumToAsc(
-              parentType === 'tuple'
+              util.isTupleType(parentType)
                 ? `this[${index}]`
                 : `this._${parentType}.${parentField}[${index}].value`,
               valueType,
@@ -261,6 +261,7 @@ module.exports = class AbiCodeGenerator {
   }
 
   _generateTupleType(inputOrOutput, index, parentClass, parentType, parentField) {
+    let type = inputOrOutput.get('type')
     let name = inputOrOutput.get('name')
     if (name === undefined || name === null || name === '') {
       name = parentType === 'event' ? `param${index}` : `value${index}`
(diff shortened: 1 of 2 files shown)


src/codegen/abi.js
```diff
@@ -594,6 +585,20 @@ module.exports = class AbiCodeGenerator {
...
+    return util.isTupleType(type)
+      ? this._tupleTypeName(inputOrOutput, index, tupleResultParentType, this.abi.name)
+      : util.isTupleArrayType(type)
+      ? `Array<${this._tupleTypeName(
+          inputOrOutput,
+          index,
+          tupleResultParentType,
+          this.abi.name,
+        )}>`
+      : typesCodegen.ascTypeForEthereum(type)
```
Jannis, line 599: Since this can be used for input and result types, let's rename the parent type argument from `tupleResultParentType` to `tupleParentType`.

src/codegen/abi.js
```diff
@@ -502,11 +489,7 @@ module.exports = class AbiCodeGenerator {
       let params = inputs.map((input, index) =>
         tsCodegen.param(
           input.get('name'),
-          input.get('type') === 'tuple'
-            ? this._tupleTypeName(input, index, tupleInputParentType, this.abi.name)
-            : util.isTupleArrayType(input.get('type'))
-              ? `Array<${this._tupleTypeName(input, index, tupleInputParentType)}>`
-              : typesCodegen.ascTypeForEthereum(input.get('type')),
+          this._getTupleParamType(input, index, tupleResultParentType),
```
Jannis, line 492: `tupleResultParentType` needs to be `tupleInputParentType` (see the old code just above this addition). `tupleResultParentType` is only used for return values, not input arguments.

Jannis:
Superseded by https://github.com/graphprotocol/graph-cli/pull/465.

PR #465, opened 2020-04-08
Fix codegen for tuple arrays

### LINKED ISSUE #455: Generated function result for tuple is of undefined type

Jannis:
Fixes #455.
Supersedes #456.

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -229,7 +229,7 @@ module.exports = class AbiCodeGenerator {
     }
 
     // Generate getters and classes for the param (classes only created for Ethereum tuple types)
-    return valueType === 'tuple'
+    return util.containsTupleType(valueType)
       ? this._generateTupleType(
           inputOrOutput,
           index,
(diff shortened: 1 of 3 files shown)


Jannis:
This addresses the review comments in #456 so we can get it into the upcoming 0.18.0 release.

JamesLefrere:
Hey Jannis, sorry I didn't get around to your comments in time - but thank you!

Jannis:
@JamesLefrere No worries, thanks a bunch for the bugfix! I'm tagging releases right now and it would be a shame for the fix not to be included, hence the update. 😄

PR #466, opened 2020-04-12
Added support of contract name in graph init command

stevepro-lab:
Currently if `graph init` is used to create project, all the files are name as `Contract`. This PR adds a optional parameter `--contract-name` in the `graph init` command. This name will be used to create subgraph project files. 

Why this is needed? 
While creating subgraph project with many contract, we use `graph init` command for each contract and merge those projects. One step while merging the projects is renaming files by contract name to make it more readable. This option will ease the process and make it less erroneous. 

Next step after this PR approach looks fine: Create separate PR to accept multiple contract addresses and name in `graph init`

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -33,11 +33,12 @@ ${chalk.dim('Options for --from-contract:')}
       --network <mainnet|kovan|rinkeby|ropsten|goerli|poa-core>
                                 Selects the network the contract is deployed to
       --index-events            Index contract events as entities
+      --contract-name           Name of the contract (default: Contract)      
 `
 
 const processInitForm = async (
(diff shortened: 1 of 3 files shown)


Jannis:
@stevepro-lab I love this, thanks for working on it! Tested it locally and it works fine. I'd like the data source name to also be changed according to the contract name. Once that's done, we'll merge the change.

Could you also change the commit message from
```
src/commands/init, src/scaffold Added support of contract name in graph init command
```
to
```
commands/init, scaffold: Add contract name support to graph init
```
? That would be awesome.

stevepro-lab:
Thanks @Jannis  I have made both the changes.

PR #471, opened 2020-04-15
Allow path prefixes in data source and template names.

fubhy:
It's currently possible (although probably by "accident") to use / in the name of  dataSources and templates. For both, it then simply creates a sub-directory when running codegen. However, for templates it then creates invalid syntax in the templates.ts output. Everything else works fine though. I actually like this because it allows us to disambiguoate contract names for different versions of the same contract without adding weird version name suffixes to the name. A simple fix would be to add a path.basename to the code generator for templates.ts. It's a one-line fix on the script. Would you accept such a PR or is there something else tied to the name property in the manifest that would cause problems with the / character?

### DIFF
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -1,5 +1,4 @@
 const immutable = require('immutable')
-const path = require('path')
 
 const tsCodegen = require('./typescript')
 const typesCodegen = require('./types')
@@ -9,7 +8,10 @@ const ABI = require('../abi')
 module.exports = class AbiCodeGenerator {
   constructor(abi) {
     this.abi = abi
-    this.name = path.basename(abi.name)
+    this.name = abi.name
+      .split('/')
+      .reverse()
+      .shift()
   }
 
   generateModuleImports() {
--- a/src/codegen/abi.js
+++ b/src/codegen/abi.js
@@ -1,4 +1,5 @@
 const immutable = require('immutable')
+const path = require('path')
 
 const tsCodegen = require('./typescript')
 const typesCodegen = require('./types')
@@ -8,6 +9,7 @@ const ABI = require('../abi')
 module.exports = class AbiCodeGenerator {
   constructor(abi) {
     this.abi = abi
+    this.name = path.basename(abi.name)
   }
 
   generateModuleImports() {
@@ -317,7 +319,7 @@ module.exports = class AbiCodeGenerator {
   }
 
   _generateSmartContractClass() {
-    let klass = tsCodegen.klass(this.abi.name, {
+    let klass = tsCodegen.klass(this.name, {
       export: true,
       extends: 'ethereum.SmartContract',
     })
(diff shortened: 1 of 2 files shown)


leoyvens:
While this behavior may be convenient and it's curious that it happens to almost work, having a `/` in a data source name could possibly have consequences in Graph Node and other software that interacts with the protocol, so I'd say it should be disallowed and names should be validated to something like alphanumerical plus `_` and `-`.

We do wish to move abis to a top level field at some point, so that data source names no longer influence codegen.

fubhy:
@leoyvens Moving ABIs to a top level field is definitely nice but would not solve the problem of ambiguous contract / abi names. Complex projects like ours usually maintain multiple versions of the same contract over time as a result of new contract deployments after bug fixes / new features / etc. Naturally, we can't come up with a new name of the same contract every time we change a small detail in our contracts. However, we need to maintain data across multiple versions of these deployments. Hence, allowing some sort of prefix (maybe even in a separate yaml property) in the manifest is extremely helpful.

fubhy:
I updated the PR to also add prefixes for ABIs.

src/codegen/abi.js
```diff
@@ -8,6 +9,7 @@ const ABI = require('../abi')
 module.exports = class AbiCodeGenerator {
   constructor(abi) {
     this.abi = abi
+    this.name = path.basename(abi.name)
```
Jannis, line 12: This will break on Windows. It's been a rarity... but people use it. We may have to use regular string splitting here.

src/codegen/template.js
```diff
@@ -23,7 +24,10 @@ module.exports = class DataSourceTemplateCodeGenerator {
   _generateTemplateType() {
     let name = this.template.get('name')
 
-    let klass = tsCodegen.klass(name, { export: true, extends: 'DataSourceTemplate' })
+    let klass = tsCodegen.klass(path.basename(name), {
```
Jannis, line 27: Same here.

fubhy:
@Jannis Good catch. Switched to str.split.

Jannis:
@fubhy Let's see if we can move forward with this. There are still a few places that need updating.

https://github.com/graphprotocol/graph-cli/blob/master/src/type-generator.js#L184-L188
https://github.com/graphprotocol/graph-cli/blob/master/src/type-generator.js#L219-L224

We're implicitly assuming that if the data source / template / ABI name is `foo/bar`, that the above lines will create folders named accordingly. However, on Windows, this won't be the case. In fact, I'm not sure what Windows would do, it maybe that `foo/bar` is a valid directory name there...

A cleaner way would be to use the split data source / template / ABI names and join paths using something like
```js
let [...p, name] = abi.abi.name.split('/')
let filename = path.join(..., ...p, `${name}.ts`)
```
This is pretty dirty. An even better way could be to add an `outputFilename()` method to data sources, templates, ABIs and use that instead of duplicating the manual splitting everywhere.

saihaj:
closing due to inactivity. Please re-create this based on new structure or file an issue.

PR #478, opened 2020-04-21
Fixing path lookups in compiler script.

fubhy:
Currently, the path values passed to `asc.main` are not correctly resolving the libs directory relative to the current working directory in all cases. Also, the build directory path is not getting resolved relative to the `cwd` when adding the --output-dir option to the `build` or `deploy` commands.

### DIFF
--- a/src/compiler.js
+++ b/src/compiler.js
@@ -181,16 +181,6 @@ class Compiler {
     }
   }
 
-  _copySubgraphFile(maybeRelativeFile, sourceDir, targetDir, spinner) {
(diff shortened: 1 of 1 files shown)


fubhy:
Another related path resolution fix for copying the schema file. Currently, when specifying the schema file in the original root subgraph.yaml as `../../schema.graphql` the utility function for copying it to the build directory resolves the destination file path relatively, resulting in the schema.graphql file ending up somewhere outside of the build directory.

pi0neerpat:
I just tested and can confirm this fixes #479 using this example repo https://github.com/pi0neerpat/graph-cli-testing-example

PR #480, opened 2020-04-24
allow to use graph-cli as library

wighawag:


Jannis:
I'm torn about this. We currently have no intention of guaranteeing the stability of code internal to graph-cli.

What's the use case for the change?

wighawag:
The use case is to distribute graph for local development 
see : https://www.npmjs.com/package/eip721-subgraph

wighawag:
I guess this could be a feature of graph-cli where you can point to a npm package?

```
graph deploy test/subgraphName --npm eip721-subgraph --ipfs http://localhost:5001 --node http://127.0.0.1:8020
```

saihaj:
closing due to inactivity. I personally don't like this approach, now that we have mono repo structure I would rather prefer splitting `cli` package to `core` so that way we can provide useful internals. Please file an issue with what would be useful for you.

PR #498, opened 2020-05-11
test: Speed up `docker-compose down`

leoyvens:
This turns out to be one of the slowest steps of testing, and this gives a nice speed up.

Looks like some formatting changes went along.


Jannis:
Good catch!

PR #542, opened 2020-08-18
Add support for CLI plugins

Jannis:
This sets Graph CLI up to detect plugin packages that extend the set of commands provided by `graph`.


Jannis:
@fordN Probably, I'm not sure. We'll deal with that when it happens. 😁

PR #556, opened 2020-09-16
Emit a warning when multiple data sources have duplicate functionality

fordN:
Includes an additional check during subgraph validation to identify data sources which are functionally equivalent. 
If a subgraph is deployed with functionally equivalent data sources it will likely lead to unexpected entity data such as double counting.


fordN:
This is currently setup to emit a warning, but I'm now thinking we should fail the subgraph on this condition.  Thoughts?

PR #582, opened 2020-11-30
manifest: Add features section

leoyvens:
See https://github.com/graphprotocol/graph-node/pull/2037.


lutter (approved):
One thought: should feature be an enum to make it clearer what is allowed as a feature? It might introduce versioning dependency as new features are added, but it would also allow validating features much earlier (before `graph-node` ever sees the manifest)

PR #613, opened 2021-02-13
fix print bug when there is an error in setting up the 'graph test' e…

Melvillian:
…nvironment.

Prior to this commit, if an error occured when setting up the 'graph test' command's test environment, the error would be printed as '[Object object]', because Javascript's String interpolation does not handle Javascript Objects well. Now it prints out the full error message, allowing developers to quickly debug why 'graph test' is not working.


leoyvens (approved):
Thank you!

Jannis (approved):

PR #621, opened 2021-02-22
parameterize timeout

tilacog:
Add a `--timeout` optional command-line parameter to define how long to wait for containers (Postgres, Ganache and IPFS) to be up and ready. 

Default is 10.000 ms, the same as the original value, so this shouldn't break the previous API.


src/commands/test.js
```diff
@@ -94,6 +96,10 @@ module.exports = {
         'test',
         standaloneNode ? 'docker-compose-standalone-node.yml' : 'docker-compose.yml',
       )
+
+    // parse timeout. Defaults to 10 seconds
+    timeout = Math.abs(parseInt(timeout)) || 10000
```
leoyvens, line 101: I'm actually ok with defaulting to 100 secs, the default should be what's most likely to work. And the timeout configuration isn't really a breaking change.

leoyvens: Or 60 seconds, anyways whatever we are using for our tests.

leoyvens (approved):

PR #629, opened 2021-03-10
Added basic url auth support

juanmardefago:
Not sure why this wasn't included, but allowing the usage of the auth parameter of the URL will allow for basic authentication (user and password) if you don't want to run a JWT authentication solution, but still want some simple form of authentication when deploying.

Was there a security reason why this wasn't included?


leoyvens (approved):
Don't see a problem with this, thanks for the PR

juanmardefago:
> Don't see a problem with this, thanks for the PR

You're welcome! Happy to help ;)

PR #632, opened 2021-03-23
Init command updated to support all hosted-service compatible networks

fordN:
Add `graph init ...` support for `poa-sokol`, `xdai`, `matic`, mumbai, `fantom`, `bsc`, and `clover`.

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -30,7 +30,7 @@ ${chalk.dim('Choose mode with one of:')}
 ${chalk.dim('Options for --from-contract:')}
 
       --abi <path>              Path to the contract ABI (default: download from Etherscan)
-      --network <mainnet|kovan|rinkeby|ropsten|goerli|poa-core>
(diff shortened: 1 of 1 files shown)


leoyvens:
If Etherscan does not support the network, what happens?

fordN:
> If Etherscan does not support the network, what happens?

I believe it continues without an ABI if it can't pull it from etherscan.  In which case the user will simply have to supply it themselves. I'll double check.

leoyvens (approved):

leoyvens:
@fordN sounds good then

AthaSSiN:
Wanted to ask if there is/will be no support for fetching the ABIs of Matic/Mumbai from Blockscout. (Matic's docs)[https://docs.matic.network/docs/develop/graph/] say that the graph can scan on Blockscout as well, but for me, graph only checks on etherscan (and Matic is not available on etherscan)

PR #642, opened 2021-04-16
add support to load generated manifests

KillerByte:
in order to use the same subgraph for multiple neworks (ex. mainnet, kovan, mainnet-ovm, etc.), it is currently
necessary to construct a fresh subgraph manifest for each network and supply them individually.

this PR is a slight modification to the CLI in order to resolve the manifest definition from a JS file. as long
as the javascript resolves its `module.exports` to the JSON of the manifest, custom code can be used to generate
the appropriate manifest

this should not break any existing compatibility with YAML or JSON loading.

### DIFF
--- a/src/migrations/util/load-manifest.js
+++ b/src/migrations/util/load-manifest.js
@@ -1,6 +1,6 @@
 const fs = require('fs-extra')
 const path = require('path')
-const yaml = require('yaml')
+const yaml = require('js-yaml')
 
 function loadManifest(manifestFile) {
   if(manifestFile.match(/.js$/)) {
--- a/src/migrations/util/load-manifest.js
+++ b/src/migrations/util/load-manifest.js
@@ -1,0 +1,15 @@
+const fs = require('fs-extra')
+const path = require('path')
+
+function loadManifest(manifestFile) {
+  try {
+    return require(path.resolve(manifestFile))
+  }
+  catch(_) {
+    return yaml.safeLoad(fs.readFileSync(manifestFile, 'utf-8'))
+  }
+}
+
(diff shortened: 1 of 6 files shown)


src/migrations/util/load-manifest.js
```diff
@@ -0,0 +1,15 @@
+const fs = require('fs-extra')
+const path = require('path')
+
+function loadManifest(manifestFile) {
+  try {
+    return require(path.resolve(manifestFile))
+  }
+  catch(_) {
+    return yaml.safeLoad(fs.readFileSync(manifestFile, 'utf-8'))
```
fordN, line 9: Looks like we'll need to import `yaml` to be used here?

fordN (changes requested):
Left a comment about loading the `yaml` module which looks like the culprit of the failed builds.  I'd also love to see a few validation tests for JS manifests loaded using this new manifest file type support. (https://github.com/graphprotocol/graph-cli/tree/master/tests/cli/validation)

KillerByte:
@fordN regarding validation tests, I am scared of adding such a test since the errors for this change would be complteely generated from Node.js, which might evolve its error format over time. Let me know what kind of tests you want to see.

also, I added a change to make the initial load of the module more intelligent, and instead it will load based on the file extension so taht the require error is not shadowed.

KillerByte:
@fordN JSYK, a test has been successfully completed on my local graph node to verify that manifests using this method are deployed and operating as expected.

fordN (approved):
Looking good! 👍 

I've tested the update as well by using deploying a test .js manifest subgraph to the hosted-service.

PR #673, opened 2021-06-02
commands/init: Add arbitrum-one (production) to networks arg

schmidsi:


leoyvens (approved):

schmidsi:
Build is failing. I'll investigate. Any hints? I can reproduce the error on my local machine, but I can not reproduce the non-error: If I checkout an old commit which had no build error, I still get the same build error 🤯

PR #674, opened 2021-06-03
codegen: Remove dead code

otaviopace:
I wonder if this should be removed as well: https://github.com/graphprotocol/graph-cli/blob/master/src/validation/manifest.js#L54


leoyvens (approved):
We don't support graphql union types, though in theory we might someday.

PR #676, opened 2021-06-04
Fix build

schmidsi:
This PR fixes the build which is failing for some days. I narrowed down the problem to the way we referenced `assemblyscript`. I know, a general update of `assemblyscript` is on the way, but this also caused problems for users: https://github.com/graphprotocol/graph-ts/issues/113#issuecomment-854811942 and several messages on Discord.

The fix is mainly just referencing a tag instead of a commit-hash and in the proper form (`git+`-prefixed):
```
"assemblyscript": "git+https://github.com/AssemblyScript/assemblyscript.git#v0.6"
```

We should also release a `v0.20.1` soon, since the current version is causing errors.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,15 +6,16 @@ language: node_js
 node_js:
   - '10'
   - '11'
-cache: yarn
+cache:
+  yarn: false
 services:
   - docker
 # before_install:
 #  - sudo apt-get install libsecret-1-dev dbus gnome-keyring python-keyring python-gnomekeyring
 # before_script:
-  # Create a new 'login' keyring; this appearas to be necessary for gnome-keyring to auto-launch
-  # properly when needed
-  # - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
+# Create a new 'login' keyring; this appearas to be necessary for gnome-keyring to auto-launch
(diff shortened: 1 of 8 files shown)


otaviopace:
The CI problem was only because of the `assemblyscript` version? Why did it start happening? I mean, the commit is still [there](https://github.com/AssemblyScript/assemblyscript/commit/36040d5b5312f19a025782b5e36663823494c2f3).

schmidsi:
> The CI problem was only because of the `assemblyscript` version? Why did it start happening? I mean, the commit is still [there](https://github.com/AssemblyScript/assemblyscript/commit/36040d5b5312f19a025782b5e36663823494c2f3).

I wasn't able to identify why it started to fail. But I was able narrow down the problem outside of this repo: When I checked out a commit with a successful build on my machine, I saw the same error. Even after `rm -rf node_modules`.

I have one assumption: Yarn or Github changed something on their hosted service which breaks the install. The reference was not standard: It missed the `git+` prefix.

If you do not see any problems with this commit, it would be great to move forward because our [builds are failing](https://travis-ci.com/github/graphprotocol/graph-cli) and [users are having problems with this release](https://discord.com/channels/438038660412342282/791444347823980604/851014429448732712).

otaviopace (approved):

PR #690, opened 2021-06-15
Update AssemblyScript to 0.19.10

otaviopace:
This PR:

- Updates AssemblyScript dependency on `package.json`
- Updates codegen for the newer version
- Adds migration changes for apiVersion on subgraphs

What's missing:
- throw error when subgraph's `apiVersion` isn't the one with the newest `AssemblyScript` version

Related PRs:

- https://github.com/graphprotocol/graph-ts/pull/185
- https://github.com/graphprotocol/graph-node/pull/2571


otaviopace:
I'll fix the tests soon, it seems that's because it's pointing to the wrong `graph-ts`. Let me just open that PR first 🙂

schmidsi (approved):
I'm not very deep into this codebase, but it looks good. Can't wait for this update actually.

src/compiler/asc.js
```diff
@@ -0,0 +1,71 @@
+const asc = require('assemblyscript/cli/asc')
+
+const createExitHandler = inputFile => () => {
+  throw new Error(`The AssemblyScript compiler crashed when compiling this file: '${inputFile}'
+Unfortunately this error has not been mapped yet.
```
leoyvens, line 5: This error message is really nice and gentle! I'd remove this part:
> Unfortunately this error has not been mapped yet

It's not clear what it means and doesn't add much to the already long message.

leoyvens (changes requested):
Is there anywhere where we validate the api version? The migration is nice but it's not guaranteed to work, we should have an error if we try to build with an api version < 0.0.5.

src/compiler/asc.js
otaviopace: Ok, I'll do it, thanks!

otaviopace:
@leoyvens that's a great question (blocking apiVersion), I'll take a look where I can put this to be guaranteed to fail

src/compiler/asc.js
otaviopace: Fixed!

leoyvens (approved):

PR #707, opened 2021-07-19
Added missing network options

promaty:


src/commands/init.js
```diff
@@ -15,6 +15,33 @@ const { chooseNodeUrl } = require('../command-helpers/node')
 const { abiEvents, generateScaffold, writeScaffold } = require('../scaffold')
 const ABI = require('../abi')
 
+const networkChoices = [
```
otaviopace, line 18: Shouldn't `arbitrum-testnet-v5` be added to this list as well?

promaty: I only see these two from the graph node metrics, I think that one doesn't exist?
```
arbitrum-one
arbitrum-rinkeby
```

otaviopace: oh okay, I was following the previous values in the constant

otaviopace (approved):

PR #727, opened 2021-08-20
add support for `hardhat` example

fernandomg:
This PR adds changes to the `--from-example` flag.

Now `--from-example` will receive two possible values `truffle` or `hardhat`.

It defaults to `truffle`, so it's backward compatible with the Truffle/Ganache "Quick Start" guide.


schmidsi:
@fernandomg Please rebase to the latest graph-cli version.

PR #728, opened 2021-08-25
codegen: Set non-nullable values in constructor with zero/empty values

otaviopace:
This is being done because if an `Entity` field (that is non-nullable) is accessed before it's set, like this:
```graphql
type Total @entity {
  id: ID!
  amount: BigInt
}
```
```tsx
let total = new Total("latest")
total.amount = total.amount + BigInt.fromI32(1);
```

It would give this runtime error:

スクリーンショット 2021-08-24 21 55 57

Because of this line of code in `src/generated/schema.ts`:

スクリーンショット 2021-08-24 21 56 51

If we set default/zero/empty values in the constructor, that can't happen no more.

Related docs:  https://gist.github.com/otaviopace/8406cb39644d2e7678570d1e7f50dac4#value-initialization


src/codegen/types/conversions.js
```diff
@@ -188,27 +188,28 @@ const VALUE_TO_ASSEMBLYSCRIPT = [
...
-  ['string', /.*/, code => `Value.fromString(${code})`],
+  ['Address', 'Bytes', code => `Value.fromBytes(${code})`, 'Address.zero()'],
+  ['Bytes', 'Bytes', code => `Value.fromBytes(${code})`, 'Bytes.empty()'],
+  ['boolean', 'Boolean', code => `Value.fromBoolean(${code})`, 'false'],
+  ['i32', 'Int', code => `Value.fromI32(${code})`, '0'],
+  ['BigInt', 'BigInt', code => `Value.fromBigInt(${code})`, 'BigInt.zero()'],
+  ['string', 'String', code => `Value.fromString(${code})`, "''"],
+  ['string', 'ID', code => `Value.fromString(${code})`, "''"],
+  ['BigDecimal', 'BigDecimal', code => `Value.fromBigDecimal(${code})`, 'BigDecimal.zero()'],
+  ['string', /.*/, code => `Value.fromString(${code})`, "''"],
```
otaviopace, line 212: Not sure if this was the best place to put the zero values, I had a separate const for this, but it didn't seem to fit the whole `src/codegen/types/*` architecture. It was something like this:

```js
const DEFAULT_AS_VALUE_CONSTRUCTION = [
  // Arrays

  ['Array<Address>', 'new Array(0)'],
  ['Array<Bytes>', 'new Array(0)'],
  ['Array<boolean>', 'new Array(0)'],
  ['Array<i32>', 'new Array(0)'],
  ['Array<BigInt>', 'new Array(0)'],
  ['Array<string>', 'new Array(0)'],
  ['Array<BigDecimal>', 'new Array(0)'],

  // Scalar values

  ['Address', 'Address.zero()'],
  ['Bytes', 'Bytes.empty()'],
  ['boolean', 'false'],
  ['i32', '0'],
  ['BigInt', 'BigInt.zero()'],
  ['string', "''"],
  ['BigDecimal', 'BigDecimal.zero()'],
]
```

otaviopace: We can also discuss if there's any chance we could be missing a type, maybe I should use RegExp for the `Array` type for example.

But I'm a bit confident that it will work in all cases since the function I created (`initializedValueFromAsc`) is almost a copy of `valueFromAsc` which is used for generating the `setters` we have [today](https://github.com/graphprotocol/graph-cli/blob/4959e871bbf4a450b310f0dfbba7c22ce6f739a0/src/codegen/schema.js#L173)

leoyvens (approved):
Haven't reviewed the code too closely but it's great that we're doing this! It might be nice to have some more complete codegen tests that would allow us to visualize the change, but we don't have to do that now.

PR #729, opened 2021-08-25
Lock dependencies

otaviopace:
These dependencies were being used by the project, but it only worked because some of our dependencies use them.

This PR locks the version of both (I've investigated first which we're using today).


leoyvens (approved):
nice!

PR #731, opened 2021-08-27
Integrating Matchstick rust binary as a runnable command in the CLI

georg-getz:
All the deletions in test.js are because the entire previous content of the file has been moved to local.js.


otaviopace:
@georg-getz hey! I'll review your PR soon, I'll just grab some context with @azf20 here.
I might reach out in Discord as well 🙂

src/commands/test.js
```diff
@@ -1,525 +1,63 @@
+const { Binary } = require('binary-install-raw')
+const { XMLHttpRequest } = require('xmlhttprequest')
```
otaviopace, line 2: We already have `node-fetch` installed in the project, have you considered using it instead? Not sure if there's a specific reason for using this one.

src/commands/test.js
```diff
@@ -1,525 +1,63 @@
...
-      toolbox.print.info(indent('  ', ethereumLogs))
-    }
-
-    // Always print the test output
-    toolbox.print.info('')
-    toolbox.print.info('  Output')
-    toolbox.print.info('  ------')
-    toolbox.print.info('')
-    toolbox.print.info(indent('  ', result.output))
+    const url = `https://github.com/LimeChain/matchstick/releases/download/${ version }/${ platform }`;
```
otaviopace, line 46: There's no need for the spaces here though, not sure if you added for readability.
You don't have to change if you don't want, we don't have a linter yet.

src/commands/test.js
```diff
@@ -1,525 +1,63 @@
...
-      let tempComposeFile = path.join(tempdir, 'compose', 'docker-compose.yml')
+function getPlatform() {
+  const type = os.type();
+  const arch = os.arch();
 
-      // Copy the compose file to the temporary directory
-      await toolbox.filesystem.copy(composeFile, tempComposeFile)
+  if (type === 'Windows_NT' && arch === 'x64') return 'binary-windows';
+  if (type === 'Linux' && arch === 'x64') return 'binary-linux';
+  if (type === 'Darwin' && arch === 'x64') return 'binary-macos';
```
otaviopace, line 60: Maybe you could aggregate all if's related to `arch` in a single one like this, so lines are shorter:

```js
if (arch === 'x64') {
  if (type === 'Windows_NT') return 'binary-windows';
  if (type === 'Linux') return 'binary-linux';
  if (type === 'Darwin') return 'binary-macos'
}
```

This is more of prefference, you don't have to change if you don't want 🙂

src/commands/test.js
```diff
@@ -1,525 +1,63 @@
...
-    if (nodeLogs) {
-      toolbox.print.info('')
-      toolbox.print.info('  Graph node')
-      toolbox.print.info('  ----------')
-      toolbox.print.info('')
-      toolbox.print.info(indent('  ', nodeLogs))
+    const platform = getPlatform();
+    if (!version) {
+      let xmlHttp = new XMLHttpRequest();
+      xmlHttp.open('GET', 'https://api.github.com/repos/LimeChain/matchstick/releases/latest', false);
```
otaviopace, line 40: I think it's a good idea to lock the version of `matchstick`.

Nowadays for example we have `graph-ts` locked to a version because sometimes we do releases with breaking changes.

georg-getz: We talked with Leo and Adam about this problem. If we want to lock the version of graph-cli to matchstick this would mean that we would have to release a new version of graph-cli every time we release a new one for matchstick (keep in mind that we are in something like an open beta right now so severe bugs are somewhat frequent) and that's why we decided it would be better to always download the latest version so that we can freely add bug fixes and new features whenever necessary while also having the version flag available if someone wants to download an older binary and in our readme file we will have a table of compatible versions.

otaviopace: Ok!

src/commands/local.js
```diff
@@ -0,0 +1,525 @@
...
+const spawn = require('child_process').spawn
+
+const { fixParameters } = require('../command-helpers/gluegun')
+const { step, withSpinner } = require('../command-helpers/spinner')
+
+// Clean up temporary files even when an uncaught exception occurs
+tmp.setGracefulCleanup()
+
+const HELP = `
+${chalk.bold('graph test')} [options] ${chalk.bold('<test-command>')}
```
otaviopace, line 18: Also, I think this it should be `graph local` and `<local-command>`

src/commands/local.js
```diff
@@ -0,0 +1,525 @@
...
+      --standalone-node <cmd>   Use a standalone Graph Node outside Docker Compose (optional)
+      --standalone-node-args    Custom arguments to be passed to the standalone Graph Node (optional)
+      --skip-wait-for-ipfs      Don't wait for IPFS to be up at localhost:15001 (optional)
+      --skip-wait-for-ethereum  Don't wait for Ethereum to be up at localhost:18545 (optional)
+      --skip-wait-for-postgres  Don't wait for Postgres to be up at localhost:15432 (optional)
+      --timeout                 Time to wait for service containers. (optional, defaults to 120000 milliseconds)
+`
+
+module.exports = {
+  description: 'Runs tests against a Graph Node environment (using Ganache by default)',
```
otaviopace, line 36: Maybe we should add a `local` between `Runs` and `tests`.

src/commands/test.js
```diff
@@ -1,525 +1,63 @@
...
-    toolbox.print.info('  Output')
-    toolbox.print.info('  ------')
-    toolbox.print.info('')
-    toolbox.print.info(indent('  ', result.output))
+    const url = `https://github.com/LimeChain/matchstick/releases/download/${ version }/${ platform }`;
 
-    // Propagate the exit code from the test run
-    process.exitCode = result.exitCode
-  },
+    let binary = new Binary(platform, url, version);
```
otaviopace, line 48: I'm not sure if I'm looking at the wrong place, but it seems that this constructor does not receive the `version` parameter: https://github.com/EverlastingBugstopper/binary-install/blob/main/packages/binary-install/index.js#L15

otaviopace:
Also this PR's CI is failing because you need to change the tests to use `graph local` instead. If you need help on which places that is, just ask here or reach out on Discord 😉

src/commands/test.js
georg-getz: Yea, I had to change the way it worked a little so I [forked](https://github.com/georg-getz/binary-install) it.
I also added it as an npm package: binary-install-raw

otaviopace: Ohh I see, I didn't know that, thanks!

otaviopace:
One other thing that's missing is to change the `README.md` description about `graph test` and add the `local` one referencing what the `test` does nowadays 🙂

otaviopace:
About the CI failing, you should change on `.travis.yml` the line that's written:

```
  - ../../bin/graph test "yarn test"
```

To:

```
  - ../../bin/graph local "yarn test"
```

Because you've changed the `graph test` to be `graph local`. **However**, the CI pointed something interesting about the new `graph test` command, this was the output from [this build](https://app.travis-ci.com/github/graphprotocol/graph-cli/jobs/534973627):

```
$ ../../bin/graph test "yarn test"
Downloading release from https://github.com/LimeChain/matchstick/releases/download/0.1.1/binary-linux
binary-linux has been installed!
/home/travis/build/graphprotocol/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-linux: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/travis/build/graphprotocol/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-linux)
```

Is `GLIBC_2.29` a dependency for running the new `graph test` command? Should people have to install something other than `graph-cli` itself?

otaviopace:
Also, does `matchstick` support the Mac M1?

I tried running the new `graph test` locally in my M1 Mac, which I thought it would fail because of the architecture, but it downloaded the binary and got this error:

```
Downloading release from https://github.com/LimeChain/matchstick/releases/download/0.1.1/binary-macos
binary-macos has been installed!
dyld: Library not loaded: /usr/local/opt/postgresql/lib/libpq.5.dylib
  Referenced from: /Users/otaviopace/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-macos
  Reason: image not found
```

Talking with @azf20 he mentioned that a similar error would happen when `postgres` was a dependency of `matchstick`.

Out of curiosity, running `os.arch()` on Node.js version `14.16.0` returns `x64` on the Mac M1.

```
Welcome to Node.js v14.16.0.
Type ".help" for more information.
> os.arch()
'x64'
> os.type()
'Darwin'
```

georg-getz:
> Also, does `matchstick` support the Mac M1?
> 
> I tried running the new `graph test` locally in my M1 Mac, which I thought it would fail because of the architecture, but it downloaded the binary and got this error:
> 
> ```
> Downloading release from https://github.com/LimeChain/matchstick/releases/download/0.1.1/binary-macos
> binary-macos has been installed!
> dyld: Library not loaded: /usr/local/opt/postgresql/lib/libpq.5.dylib
>   Referenced from: /Users/otaviopace/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-macos
>   Reason: image not found
> ```
> 
> Talking with @azf20 he mentioned that a similar error would happen when `postgres` was a dependency of `matchstick`.
> 
> Out of curiosity, running `os.arch()` on Node.js version `14.16.0` returns `x64` on the Mac M1.
> 
> ```
> Welcome to Node.js v14.16.0.
> Type ".help" for more information.
> > os.arch()
> 'x64'
> > os.type()
> 'Darwin'
> ```
> 
> **Edit**: I just cloned `matchstick` and I've been able to compile it and run the tests locally using `cargo`

Yea, the reason for that is that `postgres` **still is** a dependency of matchstick. We had removed it some time ago hence the confusion but then we made a change in graph-node which made us dependent on it again and long story short: there will be another PR opened to graph-node soon enough 😄

georg-getz:
> About the CI failing, you should change on `.travis.yml` the line that's written:
> 
> ```
>   - ../../bin/graph test "yarn test"
> ```
> 
> To:
> 
> ```
>   - ../../bin/graph local "yarn test"
> ```
> 
> Because you've changed the `graph test` to be `graph local`. **However**, the CI pointed something interesting about the new `graph test` command, this was the output from [this build](https://app.travis-ci.com/github/graphprotocol/graph-cli/jobs/534973627):
> 
> ```
> $ ../../bin/graph test "yarn test"
> Downloading release from https://github.com/LimeChain/matchstick/releases/download/0.1.1/binary-linux
> binary-linux has been installed!
> /home/travis/build/graphprotocol/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-linux: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /home/travis/build/graphprotocol/graph-cli/node_modules/binary-install-raw/bin/0.1.1/binary-linux)
> ```
> 
> Is `GLIBC_2.29` a dependency for running the new `graph test` command? Should people have to install something other than `graph-cli` itself?

Apparently we were only supporting the latest linux version. I'll add a bit of backwards compatibility.

otaviopace:
> Yea, the reason for that is that postgres still is a dependency of matchstick. We had removed it some time ago hence the confusion but then we made a change in graph-node which made us dependent on it again and long story short: there will be another PR opened to graph-node soon enough 😄

Ok! But still, I have postgres installed locally, I'm not sure what's the problem I'm facing here on my Mac M1. Do you know?

.travis.yml
```diff
@@ -26,10 +26,10 @@ script:
   - yarn
   - ../../bin/graph codegen --debug
   - ../../bin/graph build --debug
-  # Exercise `graph test` with another example subgraph
-  - cd ../../examples/basic-event-handlers
+  # Exercise `graph local` with another example subgraph
+  - cd ../basic-event-handlers
```
otaviopace, line 30: Why this changed?

georg-getz: Well I thought since both `example-subgraph` and `basic-event-handlers` folders are in the `examples` folder there is no need to go out of the `examples` folder just to reenter it. I just did it as a micro readability improvement. Is there a drawback to doing it?

otaviopace: Oh I see, okay, no problem!

otaviopace (approved):
Nice job! Looking forward to using `matchstick` 😊 

Also, I just checked and the CI passes if you are a contributor: https://app.travis-ci.com/github/graphprotocol/graph-cli/builds/236860247.

Another thing is, isn't the `getPlatform` too specific? I mean, do you really need to compile different binaries for each of the Linux versions for example? Just curious, I don't know about `matchstick`s build process. Worried a bit as well if anyone tries to use something you guys didn't expect/map.

Can I merge this? I'll see here with the team when we can release this 🙂 
It will probably be on the `0.22.0-alpha.X` series for now. We've found a bug on the new AS update that needs further investigation before releasing the stable one (I'll look into that).

georg-getz:
> Nice job! Looking forward to using `matchstick` 😊
> 
> Also, I just checked and the CI passes if you are a contributor: https://app.travis-ci.com/github/graphprotocol/graph-cli/builds/236860247. (build of your branch, but with my push just to confirm the only problem was the `docker login`)
> 
> Another thing is, isn't the `getPlatform` too specific? I mean, do you really need to compile different binaries for each of the Linux versions for example? Just curious, I don't know about `matchstick`s build process. Worried a bit as well if anyone tries to use something you guys didn't expect/map.
> 
> Can I merge this? I'll see here with the team when we can release this 🙂
> It will probably be on the `0.22.0-alpha.X` series for now. We've found a bug on the new AS update that needs further investigation before releasing the stable one (I'll look into that next week probably).

Especially when it comes to linux we **need** the different binaries. If you remember the output of the failed travis build that still ran `graph test` instead of `graph local` it said that it didn't find GLIBC_2.29 and it turns out that the highest version you can go up to on ubuntu 18.04 (Bionic Beaver) is 2.27. I tried a couple of ways to update it but all attempts failed (I don't even know if it is possible). Having different binaries for older versions fixes this problem.

Yea, you can merge it! :)

PR #732, opened 2021-08-28
Add linter

evaporei:
The linter we had wasn't properly configured. I've ran `eslint --init` once again and changed the config file a bit.

The idea here is to:

- Have the linter indicate code improvements (unused vars, duplicate vars, useless code) **and** help with code style in the future.
- Also I indend on adding rules little by little (upcoming PRs) so it's not a big change at once, this PR only includes simple ones.


otaviopace:
There's only one rule to fix `no-undef`, but I don't know what these files do and how they get these variables declarated for them.

Once I discover this, the PR is ready to go 🙂  (if necessary, I can skip those in the linter).

スクリーンショット 2021-08-28 10 04 23

PR #739, opened 2021-09-07
codegen: Only call setter in constructors if value is not derivedFrom

otaviopace:
Fixes this bug: https://gist.github.com/otaviopace/8406cb39644d2e7678570d1e7f50dac4#gistcomment-3878941

This is because derivedFrom fields only exist in query time.
Before this commit, developers would get an error similar to this:
```
field `nftsOwned` is derived and can not be set
```

Codegen before:

Screen Shot 2021-09-07 at 13 35 25

And after the change:

Screen Shot 2021-09-07 at 13 34 30

### DIFF
--- a/src/codegen/schema.js
+++ b/src/codegen/schema.js
@@ -83,9 +83,15 @@ module.exports = class SchemaCodeGenerator {
         const type = this._typeFromGraphQl(field.get('type'))
         const isNullable = type instanceof tsCodegen.NullableType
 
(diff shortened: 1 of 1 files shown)


otaviopace:
This PR doesn't contain automated tests because we don't have any for this part of the code yet.

This will be addressed later in a future PR.

tilacog (approved):

PR #764, opened 2021-10-14
Added a check specifically for M1 CPUs to fix postgres error when running graph test

georg-getz:


otaviopace:
I'll just test this branch locally and verify/see the results of this change 🙂

src/commands/test.js
```diff
@@ -53,16 +53,19 @@ function getPlatform() {
   const type = os.type();
   const arch = os.arch();
   const release = os.release();
+  const cpuCore = os.cpus()[0];
   const majorVersion = release.substr(0, release.indexOf('.'));
 
-  if (arch === 'x64') {
+  if (arch === 'x64' || (arch === 'arm64' && cpuCore.model.includes("M1"))) {
```
otaviopace, line 59: I think it's better to add the prefix `Apple `. I say that because it may happen that a CPU has a model with `M1` in their name.

src/commands/test.js
```diff
@@ -53,16 +53,19 @@ function getPlatform() {
...
     if (type === 'Darwin') {
       if (majorVersion === '19') {
         return 'binary-macos-10.15';
       } else if (majorVersion === '18') {
         return 'binary-macos-10.14';
+      } else if (cpuCore.model.includes("M1")) {
```
otaviopace, line 65: Since you do the same `.includes` here, I would put into a variable up with the other ones in the beginning of the function.

src/commands/test.js
```diff
@@ -53,16 +53,19 @@ function getPlatform() {
   const type = os.type();
   const arch = os.arch();
   const release = os.release();
+  const cpuCore = os.cpus()[0];
   const majorVersion = release.substr(0, release.indexOf('.'));
```
otaviopace, line 57: After a second look, I just realize you can use `semver` for this, we have the package included into the project already 🙂  (https://github.com/graphprotocol/graph-cli/blob/master/package.json#L25).

I know it's not a super improvement, is more to be safe and use a tested lib that's already here.

src/commands/test.js
otaviopace: I see what you've done here in the second part of this `if` statement and I'm okay with keeping it, but I don't think it gets into this branch because of it.

Example output from my M1 Mac:
スクリーンショット 2021-10-18 20 40 11

As far as I know Node.js can only run via Intel emulation, that's why it always shows up `x64` in `os.arch()`. However, once they support it, this `||` part will probably work as well.

No need to change, just detailing this for the future 🙂

georg-getz: I have pushed the changes. It is actually already possible to have `os.arch()` return `arm64` on an M1 Mac, it happened on the mac I was testing it on.

src/commands/test.js
georg-getz: Not sure what you mean here. Isn't semver used for versioning? How would I use it to determine the cpu core model?

georg-getz: Oh, I guess you mean only the last row from those 5. I have pushed that change as well.

src/commands/test.js
otaviopace: Oh really? Awesome then 🙂

PR #770, opened 2021-10-27
Fix CI breaking on new release

otaviopace:
We have a problem in this repo that we have to follow a bad release pattern:

1. Release new version on `npm`
2. Make PR and merge it

It has to be this way because otherwise the CI would break, since it tries to install the version from `package.json` on the tests, and that wouldn't be released yet.

This PR fixes the issue by adding a new env var called `GRAPH_CLI_TESTS`. Once it's set we:

1. Remove `@graphprotocol/graph-cli` dependency from whatever `package.json` generated (or cloned) from `graph init`
2. Add a `npm link` (and `unlink` for the tests teardown) to the local `graph-cli` so that we use the branch's code for the tests

This PR is organized so that it "tells a story", in the sense that the first commit is an example of a fictional version bump to show the CI breaking, then you can see how that can be fixed with the changes listed above, and then I remove the version bump just to show it continues to work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ services:
 # - dbus-launch /usr/bin/python -c "import gnomekeyring;gnomekeyring.create_sync('login', '');"
 script:
   - echo "$DOCKERHUB_PASSWORD" | docker login -u "$DOCKERHUB_USERNAME" --password-stdin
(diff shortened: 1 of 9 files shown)


leoyvens (approved):
Change looks good, love commit structure!

PR #771, opened 2021-10-27
Add release process

otaviopace:
This PR:

- Documents the release process: so that anyone with access to our `npm` package can follow the process
- Removes the publishing scripts: they didn't work + were too specific and did all at once incorrectly

If this gets approved, I'll do the same for `graph-ts`.


leoyvens (approved):
Nice!

PR #777, opened 2021-11-09
add mumbai network getabi support 

vieiraricardo:


otaviopace:
@vieiraricardo thanks for the contribution! I see you've ran `npm version minor`, but unfortunately you'll have to remove that commit, since you can't push the tag. Also because the tag you've created locally doesn't include the new things that are in `master`.

So basically if you can make this PR just be the mumbai change that would be great 😊

PR #778, opened 2021-11-09
Added a new flag for logging additional information when running graph test and fixed an issue in the if statements

georg-getz:
Many of the changes are due to semi colons being removed.


src/commands/test.js
```diff
@@ -35,48 +37,56 @@ module.exports = {
...
+function getPlatform(logs) {
+  const type = os.type()
+  const arch = os.arch()
+  const release = os.release()
+  const cpuCore = os.cpus()[0]
+  const majorVersion = semver.major(release)
+  const isM1 = cpuCore.model.includes("Apple M1")
+
+  if (logs) {
+    console.log(`OS type: ${type}\nOS arch: ${arch}\nOS release: ${release}\nOS major version: ${majorVersion}\nCPU model: ${cpuCore.model}`)
```
otaviopace, line 68: Here you can even remove the `\n`s because the back ticks are sensitive to line breaks:

```javascript
console.log(`OS type: ${type}
OS arch: ${arch}
OS release: ${release}
OS major version: ${majorVersion}
CPU model: ${cpuCore.model}`)
```

But that's only preference, choose whatever you prefer, I'm just mentioning the possibility 🙂

georg-getz: Yeah, I tried it just now but it also factors in indentation which either means that all the logs will have weird indentation in the console or they will have 0 indentation in the code which will also look weird imo.

src/commands/test.js
```diff
@@ -35,48 +37,56 @@ module.exports = {
...
+  }
 
   if (arch === 'x64' || (arch === 'arm64' && isM1)) {
     if (type === 'Darwin') {
-      if (majorVersion === '19') {
-        return 'binary-macos-10.15';
-      } else if (majorVersion === '18') {
-        return 'binary-macos-10.14';
+      if (majorVersion === 19) {
```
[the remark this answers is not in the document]
georg-getz: Yes, it's bacause the `major` method returns a number and not a string.

otaviopace (approved):

PR #779, opened 2021-11-10
add getabi support for mumbai network

vieiraricardo:


otaviopace (approved):

otaviopace:
@vieiraricardo I just realized the URL you've used is for `testnet`. I imagine that is not guaranteed to have a contract on both networks (mainnet x testnet). I think we must change this.

vieiraricardo:
> @vieiraricardo I just realized the URL you've used is for `testnet`. I imagine that is not guaranteed to have a contract on both networks (mainnet x testnet). I think we must change this.

yes, mumbai and matic are respectively testnet and mainnet from polygon and unlike the other networks it uses different endpoints to access its api. Using only one endpoint makes the CLI not get the contract from one of the networks, as it did before my PR

PR #785, opened 2021-11-23
Local Graph node debug tool using a remote GraphQL endpoint (per LimeChain)

VIVelev:
Related to and dependant on https://github.com/graphprotocol/graph-node/pull/2995

Adds the `graph debug` command to enable easy debugging of subgraphs from the cli.

---
### Some points to discuss
Since there is a significant overlap between `graph debug` and `graph deploy`, should we abstract away some of the functionality in util functions and call them in both commands where appropriate?

### DIFF
--- a/src/commands/deploy.js
+++ b/src/commands/deploy.js
@@ -30,6 +30,7 @@ Options:
   -l  --version-label <label>   Version label used for the deployment
   -h, --help                    Show usage information
   -i, --ipfs <node>             Upload build results to an IPFS node (default: ${DEFAULT_IPFS_URL})
+      --debug-fork              ID of a remote subgraph whose store will be GraphQL queried
   -o, --output-dir <path>       Output directory for build results (default: build/)
       --skip-migrations         Skip subgraph migrations (default: false)
   -w, --watch                   Regenerate types when subgraph files change (default: false)
@@ -96,6 +97,7 @@ module.exports = {
       skipMigrations,
       w,
       watch,
+      debugFork,
     } = toolbox.parameters.options
 
     // Support both long and short option variants
@@ -189,6 +191,8 @@ module.exports = {
       return
     }
 
+    // TODO: validate that debugFork is a valid IPFS hash
+
     let protocol
     try {
       // Checks to make sure deploy doesn't run against
@@ -264,7 +268,7 @@ module.exports = {
       //       `Failed to deploy to Graph node ${requestUrl}`,
       client.request(
         'subgraph_deploy',
-        { name: subgraphName, ipfs_hash: ipfsHash, version_label: versionLabel },
+        { name: subgraphName, ipfs_hash: ipfsHash, version_label: versionLabel, debug_fork: debugFork },
         async (requestError, jsonRpcError, res) => {
           if (jsonRpcError) {
             spinner.fail(


src/commands/debug.js
```diff
@@ -0,0 +1,247 @@
...
+
+const HELP = `
+${chalk.bold('graph debug')} [options] ${chalk.bold('<subgraph-name>')} ${chalk.bold(
+  '[<subgraph-manifest>]',
+)}
+
+Options:
+
+  -e, --endpoint <url>          Graph node GraphQL endpoint synced up to "block" for the debugged subgraph
+  -b, --block <number>          Usually the block at which your subgraph fails
```
otaviopace, line 20: I would remove the `Usually the`, from my understanding this is the **exact** block at which you want to debug you subgraph.

Perhaps you meant is `The block at which your subgraph usually fails`.

src/commands/debug.js
```diff
@@ -0,0 +1,247 @@
...
+    let requestUrl = new URL(node)
+    let client = createJsonRpcClient(requestUrl)
+
+    // Exit with an error code if the client couldn't be created
+    if (!client) {
+      process.exitCode = 1
+      return
+    }
+
+    let debugSubgraph = async ipfsHash => {
```
otaviopace, line 179: I know the project isn't consistent in what uses `let` and `const`, but since this won't change it would be nice to be a constant 🙂

src/commands/debug.js
```diff
@@ -0,0 +1,247 @@
...
+            // Provide helpful advice when the subgraph has not been created yet
+            if (jsonRpcError.message.match(/subgraph name not found/)) {
+              print.info(`
+Make sure to create the subgraph first by running the following command:
+$ graph create --node ${node} ${subgraphName}`)
+            }
+            process.exitCode = 1
+          } else if (requestError) {
+            spinner.fail(`HTTP error deploying the subgraph ${requestError.code}`)
+            process.exitCode = 1
```
otaviopace, line 200: Are there any other messages besides the regular `deploy` ones that we would like to map here?

otaviopace:
> Since there is a significant overlap between graph debug and graph deploy, should we abstract away some of the functionality in util functions and call them in both commands where appropriate?

About this I'm not sure yet, I do agree with you that they're very similar commands.

Perhaps it would make more sense to have a debug mode in `graph deploy` itself instead of a separate command? 🤔

src/commands/debug.js
VIVelev: I will leave it just `The block at which your subgraph fails.`

src/commands/debug.js
VIVelev: No, not really.

VIVelev:
I totally agree, it does make sense. Although it will "pollute" the `graph deploy` command with `--endpoint` and `--block`

src/commands/debug.js
VIVelev: Probably change it in [deploy.js](https://github.com/graphprotocol/graph-cli/blob/341eb940d8806f212b848a745dc8b1e101f5ef87/src/commands/deploy.js#L262) too?

VIVelev: Also what do you think about `requestUrl`, `client`, `compiler`? Should we make them consts too? I don't know what the proper JS way is, but isn't const best suited for global vars, and declaring a local as const is a bit too verbose? Of course, there may be cases where using const for special values like arrow functions is beneficial.

otaviopace (approved):

otaviopace:
@VIVelev can you rebase this with `master` please? 🙂

PR #791, opened 2021-12-03
codegen: Remove default values for primitive schema fields

otaviopace:
In the AssemblyScript update a new feature was added to the CLI.

Schemas with types such as:

```graphql
type Total @entity {
  id: ID!
  amount: BigInt
}
```

Started to have their setters called on the constructor with the default
value for the type, eg:

```typescript
export class Total extends Entity {
  constructor(id: string) {
    super();
    this.set("id", Value.fromString(id));

    this.set("amount", Value.fromBigInt(BigInt.zero()));
  }
  // ...
}
```

The problem is that these new set calls were being called with
primitives as well, that is a bug.

This commit fixes the issue by checking if the field is primitive to
avoid generating the wrong call.

### DIFF
--- a/src/codegen/schema.js
+++ b/src/codegen/schema.js
@@ -80,17 +80,24 @@ module.exports = class SchemaCodeGenerator {
       .map(field => {
         const name = field.getIn(['name', 'value'])
         const type = this._typeFromGraphQl(field.get('type'))
+
         const isNullable = type instanceof tsCodegen.NullableType
+        const isPrimitive = type.isPrimitive && type.isPrimitive()
 
         const directives = field.get('directives')
         const isDerivedFrom = directives.some(directive => directive.getIn(['name', 'value']) === 'derivedFrom')
 
-        return { name, type, isNullable, isDerivedFrom }
+        return { name, type, isNullable, isPrimitive, isDerivedFrom }
       })
       // We only call the setter with the default value in the constructor for fields that are:
+      // - Not primitive (such as Int/i32)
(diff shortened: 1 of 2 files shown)


src/codegen/schema.js
```diff
@@ -80,17 +80,24 @@ module.exports = class SchemaCodeGenerator {
...
       // We only call the setter with the default value in the constructor for fields that are:
+      // - Not primitive (such as Int/i32)
       // - Not nullable, so that AS doesn't break when subgraph developers try to access them before a `set`
       // - Not tagged as `derivedFrom`, because they only exist in query time
-      .filter(({ isNullable, isDerivedFrom }) => !isNullable && !isDerivedFrom)
+      .filter(({
+        isNullable,
+        isPrimitive,
+        isDerivedFrom,
+      }) => !(isNullable || isPrimitive) && !isDerivedFrom)
```
otaviopace, line 100: Actually thinking about it this could probably be simplified to:

```js
!(isNullable || isPrimitive || isDerivedFrom)
```

I'll check here with my boolean algebra consultants 😆

otaviopace: Yup, same truth table 🎉

dwardu:
Thanks @otaviopace. However this change seems to be incompatible with the current graph-node release 0.25.0. If I declare an `index: Int!` field in an entity, trying to saving an instance of the entity results in “missing value for non-nullable field `index`”

otaviopace:
@dwardu you're correct, I just made the truth table over the booleans we have today.
I'm about to fix the issue, expect to see `0.25.2` being published on `npm` in a few hours today 🙂

PR #796, opened 2021-12-15
Add Docker support for unit testing with Matchstick

axiomatic-aardvark:
**What is the aim of this PR?**
This PR adds support for spinning up a Docker container and running your unit tests (with [Matchstick](https://github.com/LimeChain/matchstick)) in that container.

This is useful mainly in cases where for some reason the pre-built binaries don't work for the user, and generally it's a more hassle-free approach. 

The default logic is still using the pre-built binaries, but in the future we might move on to Docker entirely.


axiomatic-aardvark:
Don't worry, we will **definitely** squash and merge 😅  Sorry for so much commits!

otaviopace:
@axiomatic-aardvark no problem about the commits. One thing though, can you please remove the `package-lock.json` and `yarn.lock` file changes since you didn't seem to add any new dependencies? 🙂

src/commands/test.js
```diff
@@ -22,49 +26,96 @@ module.exports = {
...
 
-    let binary = new Binary(platform, url, version)
-    await binary.install(force)
-    datasource ? binary.run(datasource) : binary.run()
+  if (logs_opt) {
+    console.log(`Download link: ${url}`)
   }
+
+  let binary = new Binary(platform, url, version_opt)
+  force_opt ? await binary.install(true) : await binary.install(false)
```
otaviopace, line 102: Why isn't this as before, just `await binary.install(force_opt)` without the ternary?

src/commands/test.js
```diff
@@ -90,3 +141,109 @@ function getPlatform(logs) {
 
   throw new Error(`Unsupported platform: ${type} ${arch} ${majorVersion}`)
 }
+
+function runDocker(coverage_opt, datasource, version_opt, latest_version, print) {
+  // Remove binary-install-raw binaries, because docker has permission issues
+  // when building the docker images
+  fs.rmSync("node_modules/binary-install-raw/bin", { force: true, recursive: true });
```
otaviopace, line 148: Could there be a chance that this tries to remove the directory from the wrong place?

Like running the CLI from a diferent directory or something 🤔 

Since the path here is relative.

src/commands/test.js
```diff
@@ -22,49 +26,96 @@ module.exports = {
...
+    console.log(`Download link: ${url}`)
   }
+
+  let binary = new Binary(platform, url, version_opt)
+  force_opt ? await binary.install(true) : await binary.install(false)
+  let args = ""
+
+  datasource ? args = datasource : args
+  coverage_opt ? args = args + ' -c' : args
+  args !== '' ? binary.run(args.trim()) : binary.run()
```
otaviopace, line 107: I'm okay with the last ternary statement, but I don't think they make sense in the first two lines, because you're not using the value returned in the `else` case (`: args`).

I'm not the biggest fan of inline `if`s but since you're doing one liners I think that's better than the ternary one, like this:

```javascript
  if (datasource) args = datasource
  if (coverage_opt) args = args + ' -c'
  args !== '' ? binary.run(args.trim()) : binary.run()
```


What do you think?

src/commands/test.js
```diff
@@ -22,49 +26,96 @@ module.exports = {
...
     }
 
-    const platform = getPlatform(logs)
-    if (!version) {
-      let result = await fetch('https://api.github.com/repos/LimeChain/matchstick/releases/latest')
-      let json = await result.json()
-      version = json.tag_name
+    let result = await fetch('https://api.github.com/repos/LimeChain/matchstick/releases/latest')
+    let json = await result.json()
+    let latest_version = json.tag_name
```
otaviopace, line 82: I like this variable naming better than the last one 🎉

axiomatic-aardvark:
> @axiomatic-aardvark no problem about the commits. One thing though, can you please remove the `package-lock.json` and `yarn.lock` file changes since you didn't seem to add any new dependencies? 🙂

Done ✅

src/commands/test.js
```diff
@@ -90,3 +141,109 @@ function getPlatform(logs) {
...
+  if (!fs.existsSync(dir)){
+    fs.mkdirSync(dir, { recursive: true });
+  }
+
+  try {
+    fs.writeFileSync(`${dir}/Dockerfile`, dockerfile(version_opt, latest_version))
+    print.info('Successfully generated Dockerfile.');
+  } catch (error) {
+    print.info('A problem occurred while generating the Dockerfile. Please attend to the errors below:');
+    print.info(chalk.red(error));
```
otaviopace, line 161: Did you check if this looked better than `print.error`? It probably already uses `chalk.red` under the table 🕵️‍♀️

src/commands/test.js
```diff
@@ -90,3 +141,109 @@ function getPlatform(logs) {
...
+        }
+      })
+    } else {
+      // Run the container from the existing matchstick docker image
+      spawn('docker', docker_run_opts, { stdio: 'inherit' });
+    }
+  })
+}
+
+// TODO: Move these in separate file (in a function maybe)
```
otaviopace, line 217: This is a nice TODO, I don't see as a requirement, but you can definitely do it 😊 

We do it for our `docker` files that we have for the old tests (now `graph local`).

src/commands/test.js
```diff
@@ -90,3 +141,109 @@ function getPlatform(logs) {
 
   throw new Error(`Unsupported platform: ${type} ${arch} ${majorVersion}`)
 }
+
+function runDocker(coverage_opt, datasource, version_opt, latest_version, print) {
```
otaviopace, line 145: Looking at this function I think you can improve it by using the built-in `toolbox` from `gluegun` instead of `fs` and `child_process`.

It really shouldn't change behavior, the nice thing is that everything is promisified 🙂 

Take a look at this function for instance: https://github.com/graphprotocol/graph-cli/blob/master/src/commands/init.js#L556

otaviopace (changes requested):
Some of the comments I've posted are more code style suggestions.

The behavior seems correct 🙂 

Also, I think I like Rust just as you, but can you use `camelCase` convention for variable naming instead of the `snake_case` one please? hahahah

I know we don't have a linter **yet** but I intend on adding this (and migrating the project to TypeScript 🤫 ).

Maybe one day the `graph-cli` can be in Rust 🦀

src/commands/test.js
dimitrovmaksim: In the case of `graph test -f/--force <datasource>` the value of -f/--force option won't be a boolean, but the datasource name. We pass it through gluegun to extract the datasource, but the value of the `force` option remains a string, and it won't work with `wait binary.install(force_opt)` unless we overwrite it to true.

src/commands/test.js
dimitrovmaksim: The point of using `child_process.spawn()` with `{stdio: inherit}` options was to have the docker output in real time, as well as the matchstick logs colors, as if it is ran locally. I will look into it though.

src/commands/test.js
dimitrovmaksim: Good point! For now (we should note this somewhere) the docker implementation is meant to be run from the subgraph root folder.

I will look into making it possible to be ran from anywhere.

src/commands/test.js
dimitrovmaksim: Yes, `print.error` will be better. It prints the backtrace as well which I guess will be better for debugging

src/commands/test.js
dimitrovmaksim: Ok, so basically this is an issue if someone uses the graph-cli that is installed as a dependency in the subgraph's local node_modules, e.g. `demo-subgraph/node_modules/@graphprotocol/graph-cli`, and if you run `demo-subgraph/node_modules/@graphprotocol/graph-cli/bin/graph test` the matchstick binary will be installed in `demo-subgraph/node_modules/binary-install-raw/bin`, and this will be an issue for the docker. So we can either add a .dockerignore file in the root folder of the subgraph, or delete the bin folder, as we do here. If you use the globally installed graph-cli via `npm install -g @graphprotocol/graph-cli`, this won't be an issue.

src/commands/test.js
dimitrovmaksim: Looking through the toolbox docs, `toolbox.system.run()` is `a promise wrapped around node.js child-process.exec api call`, so unfortunately it won't work for us in this case. But I think we can safely replace `fs` with `toolbox.filesystem`

src/commands/test.js
otaviopace: Ohh I see your point!

src/commands/test.js
otaviopace: Ok, nice, makes sense 👍

src/commands/test.js
```diff
@@ -90,3 +154,122 @@ function getPlatform(logs) {
...
+  // Build the Dockerfile location. Defaults to ./tests/.docker if
+  // a custom testsFolder is not delcared in the subgraph.yaml
+  let dockerDir = ""
+
+  try {
+    let doc = await yaml.load(filesystem.read('subgraph.yaml', 'utf8'))
+    testsFolder = doc.testsFolder || './tests'
+    dockerDir = testsFolder.endsWith('/') ? testsFolder + '.docker' : testsFolder + '/.docker'
+  } catch (error) {
+    print.error(error)
```
otaviopace, line 181: Over the `graph-cli`'s codebase we usually print only the `message`, but perhaps this is more useful since it will show the stacktrace as well.

src/commands/test.js
```diff
@@ -2,70 +2,134 @@ const { Binary } = require('binary-install-raw')
...
+    opts.set("docker", docker || d)
+    opts.set("force", force || f)
+    opts.set("help", help || h)
+    opts.set("logs", logs || l)
+    opts.set("recompile", recompile || r)
+    opts.set("version", version || v)
+
+    // Fix if a boolean flag (e.g -c, --coverage) has an argument
+    try {
+      fixParameters(toolbox.parameters, {
```
otaviopace, line 57: I don't think this is doing anything if you're not reading the result of `fixParameters` or the `toolbox.parameters` itself.

src/commands/test.js
```diff
@@ -90,3 +154,122 @@ function getPlatform(logs) {
...
+    return
+  }
+
+  // Create the Dockerfile
+  try {
+    await filesystem.write(`${dockerDir}/Dockerfile`, dockerfile(versionOpt, latestVersion))
+    print.info('Successfully generated Dockerfile.')
+  } catch (error) {
+    print.info('A problem occurred while generating the Dockerfile. Please attend to the errors below:')
+    print.error(error)
```
otaviopace, line 191: Here too: https://github.com/graphprotocol/graph-cli/pull/796/files#r774618697

src/commands/test.js
```diff
@@ -90,3 +154,122 @@ function getPlatform(logs) {
...
+
+  // Remove binary-install-raw binaries, because docker has permission issues
+  // when building the docker images
+  await filesystem.remove("./node_modules/binary-install-raw/bin")
+
+  // Get current working directory
+  let current_folder = await filesystem.cwd()
+
+  // Build the Dockerfile location. Defaults to ./tests/.docker if
+  // a custom testsFolder is not delcared in the subgraph.yaml
```
otaviopace, line 173: Typo `delcared` -> `declared`.

src/commands/test.js
```diff
@@ -90,3 +154,122 @@ function getPlatform(logs) {
...
+      // Build a docker image. If the process has executed successfully
+      // run a container from that image.
+      spawn(
+        'docker',
+        ['build', '--no-cache', '-f', `${dockerDir}/Dockerfile`, '-t', 'matchstick', '.'],
+        { stdio: 'inherit' }
+      ).on('close', code => {
+        if (code === 0) {
+           spawn('docker', dockerRunOpts, { stdio: 'inherit' })
+        }
```
otaviopace, line 232: You don't need to handle the error here because the stdio is inherited right? Also the option here:

```javascript
{ stdio: 'inherit' }
```

Doesn't need to include `stderr` as well? 🤔

otaviopace:
@dimitrovmaksim thanks for using the `filesystem` API and refactoring the code 😊 
I've commented a couple of small things.

src/commands/test.js
dimitrovmaksim: I thought so too, but I console logged it and it seems it updates `toolbox.parameters.array`, although I'm not sure why
Screenshot 2021-12-23 at 17 02 02

src/commands/test.js
dimitrovmaksim: According to the docs:
`'inherit': equivalent to ['inherit', 'inherit', 'inherit']`
`'inherit': Pass through the corresponding stdio stream to/from the parent process. In the first three positions, this is equivalent to process.stdin, process.stdout, and process.stderr, respectively. In any other position, equivalent to 'ignore'.`

src/commands/test.js
dimitrovmaksim: And I'm assigning the datasource here after `fixParameters` from `toolbox.parameters.array[0]` https://github.com/graphprotocol/graph-cli/pull/796/files/26f58501b72dab3e03f0768ec1918244a2db7be4#diff-7b8030c07614a0f2a6ec365e75d5bacf2fd305916553c879d3f5e2832a5b643eR77

src/commands/test.js
dimitrovmaksim: Will change it to `error.message` otherwise `print.error()` will print everything

src/commands/test.js
dimitrovmaksim: It will try to execute the next `docker run` even if the build fails, and I want it to execute only if the build is successful.

otaviopace (approved):

PR #816, opened 2022-01-17
Mock Event Fixture Generator

VIVelev:
- modifies `graph codegen` to generate a new static method (`mockNew<EventName>`) that returns a mocked event for every event specified in `subgraph.yaml`
 -  this mock event can be used as a fixture to write unit tests against, using **Matchstick**
 - the mocked event is populated with real Ethereum Event Log data from a transaction from mainnet, fetched using etherscan.io
 - a config file named `fixtures.yaml` should be specified in the root folder with the following fields:
   - **blockNumber** (ex. `7347427`)
   - **contractAddr** (ex. `0x2e645469f354bb4f5c8a05b3b30a929361cf77ec`)
   - **topic0** (ex. `0x9ab3aefb2ba6dc12910ac1bce4692cf5c3c0d06cff16327c64a3ef78228b130b`)
   - **apiKey** (ex. `<some-etherscan-api-key>`)


src/protocols/ethereum/codegen/abi.js
```diff
@@ -28,6 +33,12 @@ module.exports = class AbiCodeGenerator {
         ],
         '@graphprotocol/graph-ts',
       ),
+      tsCodegen.moduleImports(
```
otaviopace, line 36: This should probably be imported conditionally.

src/protocols/ethereum/codegen/abi.js
```diff
@@ -208,6 +225,54 @@ module.exports = class AbiCodeGenerator {
             `return new ${paramsClassName}(this)`,
           ),
         )
+
+        // Fixture generation
+        try {
```
otaviopace, line 230: I'm trying to think here if there isn't a better way to put this here.

I'm not sure I like having to do synchronous operations + big conditions here for the fixtures.

Perhaps we can go forward with this so I don't block you and then I can think of a better way to architect this so that it isn't so intrusive to the regular `abi` code generation.

src/protocols/ethereum/codegen/abi.js
```diff
@@ -208,6 +225,54 @@ module.exports = class AbiCodeGenerator {
...
+              `
+              let event = changetype<${eventClassName}>(newMockEvent());
+              ${stmnts}
+              return event;
+              `,
+            )
+          )
+        } catch (e) {
+            // no fixtures
+            console.log(e)
```
otaviopace, line 273: Also I'm not sure we want to log the errors for every time, since a lot of the times the fixture file won't be there.

src/protocols/ethereum/codegen/abi.js
```diff
@@ -208,6 +225,54 @@ module.exports = class AbiCodeGenerator {
...
+          const apiKey = args['apiKey']
+          const url = `https://api.etherscan.io/api?module=logs&action=getLogs&fromBlock=${blockNumber}&toBlock=${blockNumber}&address=${contractAddr}&${topic0}=topic0&apikey=${apiKey}`;
+
+          let resp = request("GET", url)
+          let body = JSON.parse(resp.getBody("utf8"))
+          if (body.status === '0') {
+            throw new Error(body.result)
+          }
+
+          let res = web3.eth.abi.decodeLog(
```
otaviopace, line 244: I was wondering too if shouldn't we use the `decoding` library directly instead of importing the whole `web3` here.

It actually uses a smaller library under the hood.

src/protocols/ethereum/codegen/abi.js
VIVelev: This could also be made conditionally.

package.json
```diff
@@ -24,7 +24,9 @@
     "prettier": "^1.13.5",
     "request": "^2.88.0",
     "semver": "7.3.5",
+    "sync-request": "^6.1.0",
     "tmp-promise": "^3.0.2",
+    "web3-eth-abi": "^1.7.0",
```
otaviopace, line 29: Thanks!!

otaviopace (approved):
I'll make the CI then merge it 🙂

PR #820, opened 2022-01-23
add oasis chain

badangrycloud:
Implementing a support for the oasis chain


maoueh:
Is Oasis an EVM based chain?

badangrycloud:
Yes Oasis has a EVM compatibility with his emerald rpc

otaviopace:
Hello @badangrycloud thanks for the contribution! Unfortunately I don't think this should be merged yet. `oasis` is a chain that's not listed on our docs: https://thegraph.com/docs/en/developer/create-subgraph-hosted/#supported-networks.

All of those listed there are tested and pass through a beta phase. Once `oasis` gets there I'm fine with adding here as well 🙂 
Note: I'm not sure how the roadmap for new EVM compatible chains is going, but if there's demand it might be worth prioritizing.

PR #821, opened 2022-01-28
Moved to one SubgraphManifest schema file per protocol

maoueh:
This will make it easier to add new network and will improve error messages when someone will be using fields that do not exist on a specific network but does in other (`temlates` for example used in NEAR were given a bunch of errors like the support was possible while it should have displayed an error on the `templates` node itself).

Using `templates` on NEAR before this PR:

```
✔ Apply migrations
✖ Failed to load subgraph from ../graph-node-dev/subgraphs/near/data-source/testnet.yaml: Error in ../graph-node-dev/subgraphs/near/data-source/testnet.yaml:

  Path: templates > 0 > source
  No value provided

  Path: templates > 0 > mapping
  Unexpected key in map: receiptHandlers

  Path: templates > 0 > mapping > abis
  No value provided
```

And with this PR:

```
$ ./bin/graph build ../graph-node-dev/subgraphs/near/data-source/testnet.yaml
✖ Failed to load subgraph from ../graph-node-dev/subgraphs/near/data-source/testnet.yaml: Error in ../graph-node-dev/subgraphs/near/data-source/testnet.yaml:

  Path: /
  Unexpected key in map: templates
```


Jannis:
One remark I have is that I can easily see Cross-Chain subgraphs becoming a thing and then the manifest would be a mix of data sources from different protocol.

schemas/ethereum.graphql
```diff
@@ -0,0 +1,86 @@
+scalar String
```
otaviopace, line 1: What do you think about having the manifest schemas inside each of their respective `protocol` folder?

Like this:
```
src
│   ...
└─── protocols
   └─── ethereum
      └─── manifest.graphql
   └─── near
      └─── manifest.graphql

File: tests/cli/validation/invalid-manifest.stderr [hunk 2]
```diff
@@ -1,31 +1,5 @@
 - Load subgraph from subgraph.yaml
 ✖ Failed to load subgraph from subgraph.yaml: Error in subgraph.yaml:
 
-  Path: specVersion
```
otaviopace, line 4: I like this test, can't we keep both? One for when the protocol is unknown and the other for when the validation of one of them failed.

src/protocols/index.js
```diff
@@ -1,6 +1,7 @@
 const immutable = require('immutable')
 const EthereumTypeGenerator = require('./ethereum/type-generator')
 const EthereumTemplateCodeGen = require('./ethereum/codegen/template')
+const NearTemplateCodeGen = require('./near/codegen/template')
```
otaviopace, line 4: This change is making the CI fail, there's no such file.

src/protocols/index.js
```diff
@@ -118,13 +125,17 @@ module.exports = class Protocol {
...
+    }
+
     switch (this.name) {
       case 'ethereum':
         return new EthereumTemplateCodeGen(template)
       default:
-        throw new Error(
-          `Template data sources with kind '${this.name}' are not supported yet`,
-        )
+        throw new Error(`Template data sources with kind '${this.name}' is unknown`)
```
otaviopace, line 138: Noice now the error message matches more the context 👌

src/subgraph.js
```diff
@@ -39,9 +39,21 @@ const buildCombinedWarning = (filename, warnings) =>
 
 module.exports = class Subgraph {
   static async validate(data, protocol, { resolveFile }) {
+    if (protocol.name == null) {
```
otaviopace, line 42: I'm trying to recall why I didn't add this error/check in the `Protocol` class constructor, I'm almost sure there's a scenario where we can't validate the `protocol` or we don't have the name 🤔 

Perhaps I'm daydreaming, restructuring all of this code and adding the whole concept of `protocol` was an adventure in itself 😅 (very fun though)

I'll try to remember it, but I think there **might** be a case where this validation shouldn't happen.

schemas/ethereum.graphql
maoueh: I thought about it because I saw the `protocol` subfolder. If you prefer that location, I'm all good with it. Ask and you shall receive :D

src/protocols/index.js
maoueh: Spurious change that slipped through and is actually part of another PR #824, will remove that.

src/subgraph.js
maoueh: I'll wait for your memory to wake up :) I started and put in `Protocol` first, but that was actually not given a good user experience since the validation looked more like an error than a validation.

So I moved the validation to its current location, which gives a better developer experience if protocol cannot be inferred.

This was required first because of a unit test that was failing as well as it make sense, otherwise there will be no schema to load.

tests/cli/validation/invalid-manifest.stderr
maoueh: Sure I thought there would be other cases covering those, but I can definitely re-add it to be more like it was before (in addition to this current version).

src/subgraph.js
otaviopace: I think that was it, I'm happy that a test caught it!

I can run this branch once you fix the issues too just to be sure.

maoueh:
Going to update all this today or tomorrow.

tests/cli/validation.test.js
```diff
@@ -9,6 +9,14 @@ describe('Validation', () => {
       exitCode: 1,
     },
   )
+  cliTest(
```
otaviopace, line 12: Yaaay!! 🎉

otaviopace (approved):

PR #824, opened 2022-01-29
Feature/near data source template

maoueh:


src/protocols/near/codegen/template.js
```diff
@@ -0,0 +1,40 @@
+const tsCodegen = require('../../../codegen/typescript')
+
+module.exports = class NearTemplateCodeGen {
+  constructor(template) {
+    this.template = template
+  }
+
+  generateModuleImports() {
+    return []
```
otaviopace, line 9: I think you'll need to add `'near'` here.

maoueh: It's not required because there is actually no `near` type involved in the generated code:

```
import { DataSourceTemplate, DataSourceContext } from "@graphprotocol/graph-ts";

export class receipts extends DataSourceTemplate {
  static create(account: string): void {
    DataSourceTemplate.create("receipts", [account]);
  }

  static createWithContext(account: string, context: DataSourceContext): void {
    DataSourceTemplate.createWithContext("receipts", [account], context);
  }
}
```

I tested the full flow of NEAR data source template and it worked great.

otaviopace: Oh I see why I thought that, that's because you're not using the `Account` type from the `near` module. I would prefer that instead of using `string` directly. What do you think?

maoueh: I'm ambivalent. 

If we go with `Account`, it should be used thoroughly everywhere, right now it's a mixed usage in `graph-ts` (mostly 50/50).

The second thing that makes it me doubt is how people on NEAR are used to work directly with `string`. I've research this a little bit and it seems `AccountId` is used more (https://github.com/near/near-sdk-rs/blob/05c389c75a568af3f37a3346d6f7b4064852e1de/near-sdk/src/types/account_id.rs#L39) although I've seen various example dealing with String directly. The AssemblyScript however does not seems to have an AccountId alias and seems to deal with `string` directly: https://github.com/near/near-sdk-as/blob/master/sdk-core/assembly/contract.ts#L12.

I prefer `string` personally as it make the cognitive load lower. If we choose to still wrap it, I'm voting more for `AccountId` to align with the Rust SDK for NEAR.

otaviopace: Hmm I see 🤔 

Well we can discuss that later then, I'm gonna merge this, thanks for the PR!!

otaviopace:
@maoueh this is missing a conflict resolution because of the refactor you've done that got merged into `master`

PR #826, opened 2022-01-31
graph-cli test command docker adjustments

dimitrovmaksim:
Issues:
https://github.com/LimeChain/matchstick/issues/275
https://github.com/LimeChain/matchstick/issues/283

What this PR does:

1. Matchstick offers the the option for users to define a custom libs or test folder. Until now the test folder could be declared in the `subgraph.yaml` with the `testsFolder` key, but adding this key was breaking `codegen` and `build` commands. And the libs folder was passed with the `-l` flag to the matchstick binary, which conflicts with `-l/--logs` flag in `graph-cli test`. So we decided to add a separate `matchstick.yaml` config. The `graph-cli test` command previously used the subgraph.yaml to check if `testsFolder` is declared and use it to store the created `Dockerfile` there if the `-d/--docker`   flag is used, and now it will look for `matchstick.yaml` instead

2. Changes to the docker option:
    - Now it fetches the [Dockerfile](https://github.com/LimeChain/demo-subgraph/blob/main/Dockerfile) template from the demo-subgraph repo. Using `glueguns's` `patching`, replaces the version placeholder with either the latest version or the one passed with the `-v/--version` flag by the user. This way the Dockerfile can be quickly updated, without the need to make a PR to the graph-cli or release new versions.
    - Also changes in which cases the Dockerfile is [re]created. Now it's when either it does not exist, a matchstick version is passed with `-v/--version flag` or the `-f/--force` flag is passed.
    - Removes the `--no-cache` flag from the `docker build` command

### DIFF
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -2,7 +2,7 @@ const { Binary } = require('binary-install-raw')
 const os = require('os')
 const chalk = require('chalk')
 const fetch = require('node-fetch')
-const { filesystem, print } = require('gluegun')
+const { filesystem, print, patching } = require('gluegun')
 const { fixParameters } = require('../command-helpers/gluegun')
 const semver = require('semver')
 const { spawn, exec } = require('child_process')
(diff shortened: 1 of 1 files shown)


src/commands/test.js
```diff
@@ -169,27 +169,37 @@ async function runDocker(datasource, opts) {
...
-    print.info('A problem occurred while generating the Dockerfile. Please attend to the errors below:')
-    print.error(error.message)
-    return
+  // Check if the Dockerfil already exists
+  let dockerfileExists = filesystem.exists(dockerfilePath)
+
+  // Generate the Dockerfile only if it doesn't exists,
+  // version flag and/or force flag is passed.
+  if(!dockerfileExists || versionOpt || forceOpt) {
+    print.info("Generating Dockerfile...")
```
otaviopace, line 200: You can use a `gluegun.spinner` here if you want 🙂

src/commands/test.js
```diff
@@ -214,7 +224,7 @@ async function runDocker(datasource, opts) {
     // else it'll return the image ID. Skip `docker build` if an image already exists
     // If `-v/--version` is specified, delete current image(if any) and rebuild.
     // Use spawn() and {stdio: 'inherit'} so we can see the logs in real time.
-    if(stdout === '' || versionOpt || forceOpt) {
+    if(!dockerfileExists || stdout === '' || versionOpt || forceOpt) {
```
otaviopace, line 227: Why if the `Dockerfile` exists it shouldn't remove the image?

src/commands/test.js
```diff
@@ -169,27 +169,37 @@ async function runDocker(datasource, opts) {
...
-    print.error(error.message)
-    return
+  // Check if matchstick.yaml config exists
+  if(filesystem.exists('matchstick.yaml')) {
+    try {
+      // Load the config
+      let config = await yaml.load(filesystem.read('matchstick.yaml', 'utf8'))
+
+      // Check if matchstick.yaml is not empty
+      if(config != undefined) {
```
otaviopace, line 182: I think it's better to use `config != null` cause it tests both against `undefined` and `null`.

JS is weird.

src/commands/test.js
```diff
@@ -169,27 +169,37 @@ async function runDocker(datasource, opts) {
...
+        // If a custom tests folder is declared update dockerfilePath
+        testsFolder = config.testsFolder || './tests'
```
otaviopace, line 184: Do you need `./` in `./tests` here?

If not I would recommend using Node.js's `path.join`, you can remove the ternary below:

スクリーンショット 2022-02-01 16 57 12

src/commands/test.js
dimitrovmaksim: I will check it out

src/commands/test.js
dimitrovmaksim: Here the idea is that it should build the image if the Dockerfile did not exists at the time the `graph test -d` command was run, an image does not exists, or the rebuild is forced by passing a matchstick version or the force tag.

otaviopace (approved):

PR #827, opened 2022-02-01
Tendermint integration

pienkowb:


manifest-schema.graphql
```diff
@@ -8,7 +8,8 @@ type Schema {
```
otaviopace, line 8: You'll have to rebase this with `master`, @maoueh just refactored this

src/protocols/index.js
```diff
@@ -64,6 +67,9 @@ module.exports = class Protocol {
         'near-mainnet',
         'near-testnet'
       ],
+      tendermint: [
+        'cosmoshub-4'
```
otaviopace, line 71: I'm curious what this means, ELI5 👶

src/protocols/tendermint/contract.js
```diff
@@ -0,0 +1,12 @@
+module.exports = class TendermintContract {
+  static identifierName() {
+    return 'address'
```
otaviopace, line 3: Hmmm perhaps we should change this, from my understanding we'll have "chains" instead of "contracts".

src/protocols/tendermint/contract.js
```diff
@@ -0,0 +1,12 @@
...
+  }
+
+  validate() {
+    return {
+      valid: true,
+      error: "",
```
otaviopace, line 9: This should be `null` when validation goes right.

src/protocols/tendermint/contract.js
```diff
@@ -0,0 +1,12 @@
+module.exports = class TendermintContract {
```
otaviopace, line 1: I think you can delete this file 🙂 

It's only required if you call `validateContractValues` in the `validation` step, but it's empty in `tendermint`'s `subgraph.js` file (`immutable.List()`).

otaviopace:
I'm actually kind of surprised you had to do so little code to integrate with `tendermint`, it feels like something is missing (did I do the right abstractions here so it's easy? 😅 ).

Well, when chatting with @azf20 we were talking if it would be possible to fetch the "chain" types for a subgraph just like we do the `ABI` for EVM compat subgraphs.

So then you could send that on deployment to `graph-node` and they could be used. 🙂 🎉 

(we can discuss this in person)

roignpar:
I tried rebasing with latest master, but it turned out to be a pain, so I manually redid (and adapted) all the changes on top of latest master.

src/protocols/tendermint/contract.js
otaviopace, line 3: We should fix this later, since Tendermint doesn't have `addresses`.

otaviopace (approved):

otaviopace:
I'll make the CI run just to be sure then we can merge this 😊

PR #828, opened 2022-02-03
Graph test command: Fix Linux version

dimitrovmaksim:
What it does:

The `os` module can't determine the Linux distro and version correctly, so until now the `graph test` command always downloaded the linux-20 binary. Now if the OS type is Linux, it will grep the content of `/etc/*-release` and fetch the distro name and full version and use them to determine which linux binary to download, and will display better info when `-l` flag is passed.

### DIFF
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -126,10 +126,10 @@ async function getPlatform(logsOpt) {
   const arch = os.arch()
   const cpuCore = os.cpus()[0]
   const isM1 = cpuCore.model.includes("Apple M1")
-  const linuxInfo = type === 'Linux' ? await getLinuxInfo() : []
-  const linuxDistro = linuxInfo[0] || type
-  const release = linuxInfo[1] || os.release()
-  const majorVersion = parseInt(linuxInfo[1]) || semver.major(release)
+  const linuxInfo = type === 'Linux' ? await getLinuxInfo() : new Map()
+  const linuxDistro = linuxInfo.get('name')
+  const release = linuxInfo.get('version') || os.release()
+  const majorVersion = parseInt(linuxInfo.get('version'), 10) || semver.major(release)
 
   if (logsOpt) {
     print.info(`OS: ${linuxDistro || type}\nOS arch: ${arch}\nOS release: ${release}\nOS major version: ${majorVersion}\nCPU model: ${cpuCore.model}`)
@@ -161,8 +161,15 @@ async function getPlatform(logsOpt) {
 
 async function getLinuxInfo() {
   try {
-    let info = await system.run("cat /etc/*-release | grep -E '(^VERSION|^NAME)='", {trim: true})
-    return info.replace(/[VERSION=]|[NAME=]|['"]+/g, '').split('\n')
+    let result = await system.run("cat /etc/*-release | grep -E '(^VERSION|^NAME)='", {trim: true})
+    let infoArray = result.replace(/['"]+/g, '').split('\n').map(p => p.split('='))
+    let infoMap = new Map();
+
+    infoArray.forEach((val) => {
+      infoMap.set(val[0].toLowerCase(), val[1])
+    });
+    print.info(infoMap)
+    return infoMap
   } catch (error) {
     print.error(`Error fetching the Linux version:\n ${error}`)
     process.exit(1)
--- a/src/commands/test.js
+++ b/src/commands/test.js
@@ -124,13 +124,15 @@ async function runBinary(datasource, opts) {
 async function getPlatform(logsOpt) {
   const type = os.type()
   const arch = os.arch()
-  const release = os.release()
   const cpuCore = os.cpus()[0]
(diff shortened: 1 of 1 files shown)


src/commands/test.js
```diff
@@ -156,6 +159,16 @@ function getPlatform(logsOpt) {
   throw new Error(`Unsupported platform: ${type} ${arch} ${majorVersion}`)
 }
 
+async function getLinuxInfo() {
+  try {
+    let info = await system.run("cat /etc/*-release | egrep -e '(^VERSION|^NAME)='", {trim: true})
```
otaviopace, line 164: I never used `egrep`, can we assume it's installed in most Linux computers? (it seems I have on my M1 Mac, not that it's related haha)

dimitrovmaksim: I think egrep is a shortcut for `grep -E`

dimitrovmaksim: Oh it seems it's deprecated, I'll change it

src/commands/test.js
```diff
@@ -156,6 +159,16 @@ function getPlatform(logsOpt) {
...
+    return info.replace(/[VERSION=]|[NAME=]|['"]+/g, '').split('\n')
```
otaviopace, line 165: What do you think of returning those values as an object instead of an array? I think it will be clearer to understand what's each value (instead of relying on indexes).

```javascript
[ 'Ubuntu', '20.04.1 LT (Focal Fossa)' ]
```

vs
```javascript
{ name: 'Ubuntu', version: '20.04.1 LT (Focal Fossa)' }
```

src/commands/test.js
```diff
@@ -121,16 +121,18 @@ async function runBinary(datasource, opts) {
...
-  const majorVersion = semver.major(release)
   const isM1 = cpuCore.model.includes("Apple M1")
+  const linuxInfo = type === 'Linux' ? await getLinuxInfo() : []
+  const linuxDestro = linuxInfo[0] || type
+  const release = linuxInfo[1] || os.release()
+  const majorVersion = parseInt(linuxInfo[1]) || semver.major(release)
 
   if (logsOpt) {
-    print.info(`OS type: ${type}\nOS arch: ${arch}\nOS release: ${release}\nOS major version: ${majorVersion}\nCPU model: ${cpuCore.model}`)
+    print.info(`OS: ${linuxDestro || type}\nOS arch: ${arch}\nOS release: ${release}\nOS major version: ${majorVersion}\nCPU model: ${cpuCore.model}`)
```
otaviopace, line 135: I don't want to force a refactor into the function, but maybe the `||` here could be in the value `type` itself, but as mentioned, this would require moving somethings around, Iet that decision to you 🙂

src/commands/test.js
```diff
@@ -121,16 +121,18 @@ async function runBinary(datasource, opts) {
...
   const type = os.type()
   const arch = os.arch()
-  const release = os.release()
   const cpuCore = os.cpus()[0]
-  const majorVersion = semver.major(release)
   const isM1 = cpuCore.model.includes("Apple M1")
+  const linuxInfo = type === 'Linux' ? await getLinuxInfo() : []
+  const linuxDestro = linuxInfo[0] || type
+  const release = linuxInfo[1] || os.release()
+  const majorVersion = parseInt(linuxInfo[1]) || semver.major(release)
```
otaviopace, line 132: Not sure if it's still a good practice, but I remember that it's a "good practice" to pass the radix (10 in this case) argument to `parseInt`, because JS tries to figure it out in some cases.

But I guess that might be more of a problem in browsers? Well, I will just note here if we have issues in the future, it might be this.

otaviopace: Hmmm https://stackoverflow.com/questions/44971276/why-is-it-recommended-to-provide-optional-radix-parameter-to-parseint

src/commands/test.js
dimitrovmaksim: `type` is used later and I don't want to overwrite it with the linux distro name, although the `const linuxDistro = linuxInfo[0] || type` currently does not make sense atm

src/commands/test.js
dimitrovmaksim: Actually thinking about it, I don't know if the order of those line will be always the same, maybe I should grep them separately

src/commands/test.js
```diff
@@ -156,6 +159,23 @@ function getPlatform(logsOpt) {
...
+async function getLinuxInfo() {
+  try {
+    let result = await system.run("cat /etc/*-release | grep -E '(^VERSION|^NAME)='", {trim: true})
+    let infoArray = result.replace(/['"]+/g, '').split('\n').map(p => p.split('='))
+    let infoMap = new Map();
+
+    infoArray.forEach((val) => {
+      infoMap.set(val[0].toLowerCase(), val[1])
+    });
+    print.info(infoMap)
```
otaviopace, line 171: Should this print be kept here?

otaviopace (approved):

PR #830, opened 2022-02-08
Graph test: Match against "Apple processor" for M1 chips

dimitrovmaksim:
What this PR does:

In some cases `os` module reports M1 chips as `Apple processor` instead of `Apple M1`. With this change, when running the `graph test` command, it will now look for both `Apple M1` and `Apple processor`


src/commands/test.js
```diff
@@ -125,7 +125,7 @@ async function getPlatform(logsOpt) {
   const type = os.type()
   const arch = os.arch()
   const cpuCore = os.cpus()[0]
-  const isM1 = cpuCore.model.includes("Apple M1")
+  const isM1 = /Apple M1|processor/i.test(cpuCore.model)
```
otaviopace, line 128: The new M1s return this differently? 🤔 
Mine is the 13' M1.
スクリーンショット 2022-02-08 7 20 00

otaviopace (approved):

src/commands/test.js
dimitrovmaksim: I'm not sure which one tbh, a discord user reported this:
image
It seems he is running an earlier version of BigSur

maoueh: Can you ask him to run `node -e 'console.log(os.cpus())'` on his machine and report the result.

Also the `regex` `Apple M1|processor` is too broad, would match `AMD processor` for example. I imagine you meant: `Apple (M1|processor)`.

```
$ node -e 'console.log(/Apple M1|processor/.test("AMD processor"))'
true

$ node -e 'console.log(/Apple M1|processor/.test("Intel processor"))'
true

$ node -e 'console.log(/Apple (M1|processor)/.test("Intel processor"))'
false
```

dimitrovmaksim:
```
[
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 14283310, nice: 0, sys: 12635820, idle: 123965930, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 13065070, nice: 0, sys: 11087350, idle: 126721260, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 9884450, nice: 0, sys: 7660260, idle: 133328810, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 13635780, nice: 0, sys: 4159660, idle: 133077930, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 18746170, nice: 0, sys: 5749500, idle: 126377740, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 11083520, nice: 0, sys: 9001350, idle: 130788710, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 6328300, nice: 0, sys: 1648910, idle: 142896050, irq: 0 }
  },
  {
    model: 'Apple processor',
    speed: 24,
    times: { user: 9456370, nice: 0, sys: 2563550, idle: 138853390, irq: 0 }
  }
]
```

dimitrovmaksim: Also said that `About This Mac` reports it as `Apple M1`

maoueh: Ok, the only thing I fear is that we get false positive now. Maybe we should also check the architecture. For an M1 it should always been `arm64`. So what I would do as the logic:

```
isM1 = cpu starts with Apple M1 || cpu starts with Apple processor && architecture == arm64
```

dimitrovmaksim: I think this like handles it https://github.com/graphprotocol/graph-cli/blob/0ebac1374c72252dd775a81b3626e6e3914655d1/src/commands/test.js#L138, although it this case `isM1` should be something like `isAppleChip`

dimitrovmaksim: Moved the `arch` check from the `if` on line 138 to the `isM1` declaration, I think that should cover it.

maoueh: I'm good with it.

PR #835, opened 2022-02-21
deps: Use exact versions (lock) to avoid unwanted updates

otaviopace:
This is being done to avoid unwanted updates from maintainers, eg: [`colors` incident](https://github.com/graphprotocol/graph-cli/pull/806#issue-1098110092).

Also I've separated each dependency update into a separate commit to:

- Ease review (so you can see what each version bump separately)
- Because I had issues when I tried to do all on one commit (one of the dependency resolutions went wrong because we needed multiple versions of it, so it was easier to separate it all to catch the bug)

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,7 +38,7 @@
     "jest": "26.0.0",
     "spawn-command": "0.0.2-1",
     "strip-ansi": "6.0.0",
-    "tern": "^0.24.0"
+    "tern": "0.24.3"
   },
   "scripts": {
     "test": "jest --verbose",
--- a/package.json
+++ b/package.json
@@ -35,9 +35,9 @@
   },
   "devDependencies": {
     "eslint": "6.8.0",
-    "jest": "^26.0.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -6,39 +6,39 @@
...
-    "fs-extra": "^9.0.0",
-    "glob": "^7.1.2",
+    "chalk": "3.0.0",
+    "chokidar": "3.0.2",
+    "debug": "4.1.1",
+    "docker-compose": "0.23.2",
+    "dockerode": "2.5.8",
+    "fs-extra": "9.0.0",
+    "glob": "7.1.2",
     "gluegun": "https://github.com/edgeandnode/gluegun#v4.3.1-pin-colors-dep",
```
otaviopace, line 16: I'll update `gluegun` in another PR, since they've added my fix in the latest version but I need to check if doing a big version jump breaks `graph-cli`.

tilacog (approved):

Issue #837, opened 2022-02-22
Network configuration file


azf20:
While I am working on a subgraph, I may want to update its contracts or networks:
- During local development, if I deploy a new version of the contract, and want the subgraph to index that newly deployed version
- If the protocol is deployed on multiple chains, I may want to redeploy the same subgraph on a different chain, with the addresses and startBlock for the contracts on that chain

*Currently they have to update the `subgraph.yaml` by hand, or develop their own solutions - indeed they are [directed to do so in The Graph's documentation](https://thegraph.com/docs/en/hosted-service/deploy-subgraph-hosted/#deploying-the-subgraph-to-multiple-ethereum-networks)*

**We should allow subgraph developers to easily update their subgraph configuration for a network during development using `graph-cli`**

**Proposal**: Create a simple `networks.json` standard, which can be referenced during `graph build`

`networks.json` specification:
```json
{
    "mainnet": { // the network name
        "dataSource1": { // the dataSource name
            "address": "0xabc...", // the contract address
            "startBlock": 123456 // the startBlock
        },
        "dataSource2": {
            "address": "0x123...",
            "startBlock": 123444
        }
    },
    "optimism": {
        "dataSource1": {
            "address": "0x987...",
            "startBlock": 123
        },
        "dataSource2": {
            "address": "0xxyz..",
            "startBlock": 123
        }
    }
        
}  
```

We can then add `--network` as an option in `graph build`
```bash
graph build [options] [<subgraph-manifest>]

New options:

      --network <name>        Network to use from the networkFile
      --networkFile <path>    Network file (default: "./networks.json")
```

When running `graph build`:
- If a `--network` is specified:
    - `graph-cli` should iterate through the dataSources in the specified `<subgraph-manifest>`, updating the `network`, `address` and `startBlock` based on the specified `network` from the `networkFile`
        - There will need to be slightly different treatment for different `protocols` (e.g. NEAR has `accounts` rather than `addresses`)
    - Then the existing `graph build` function should be called
- If any dataSources in the subgraph manifest are not updated, this should be logged by `graph-cli`
- If any dataSources in the configuration file are not used, this should be logged by `graph-cli`

When running `graph init`, a `networks.json` should be auto-generated based on the provided information. *Users will then be able to easily update or add additional networks.*

[Reference](https://hackmd.io/Fc0gQXHDT8-OtqVf76e23w?view)

PR #848, opened 2022-03-10
Add network config file

dimitrovmaksim:
Resolves https://github.com/graphprotocol/graph-cli/issues/837

What this PR does:

1.  `graph init` will now also generate `networks.json` from the data in `subgraph.yaml` file in the format specified in the linked Issue.

2. `graph build`:
  - the command now accepts two more options `--network <name>` and `--network-file <path>`
  - passing the `--network <name>` option will update the `network`, `address/account` (NEAR uses `account`) and `startBlock` for each dataSources in subgraph.yaml to the values defined in the networks.json file. Also automatically updates all templates' (if any) network to the one passed with `--network` option. After that the build command will continue with the existing `build` functionality.
  - throws an error if networks file does not exist
  - throws an error if network not specified in networks file
  - throws an error if a given dataSource isn't in the networks.json for the specified network.
  - logs If any dataSources in the subgraph manifest are not updated
  - logs If any dataSources in the configuration file are not used
 
Issues:
When running yarn test, jest does not exit, but if I run every test suite separately it exits correctly

```
Jest has detected the following 1 open handle potentially keeping Jest from exiting:

  ●  PROCESSWRAP

      at Process.emitInitNative (node:internal/async_hooks:205:43)
      at new ChildProcess (node:internal/child_process:263:18)
      at spawn (node:child_process:695:17)
      at start (node_modules/sync-rpc/lib/index.js:33:13)
      at sendMessage (node_modules/sync-rpc/lib/index.js:133:17)
      at createClient (node_modules/sync-rpc/lib/index.js:173:27)
      at Object.<anonymous> (node_modules/sync-request/lib/index.js:16:14)
```


otaviopace:
I just pushed a copy of your branch to trigger the CI.

I'll try to get write access to you LimeChain folks soon 🙂

src/command-helpers/network.js
```diff
@@ -0,0 +1,135 @@
...
+    `Warnings while updating sources network`,
+    async spinner => {
+      let allNetworks
+
+      try {
+        step(spinner, `Reading networks config`)
+        allNetworks = await toolbox.filesystem.read(networksFile, "json")
+      }  catch (error) {
+        throw new Error(error.message)
+      }
```
otaviopace, line 18: Is this `try`/`catch` really needed?

src/command-helpers/network.js
```diff
@@ -0,0 +1,135 @@
...
+// Checks if any network attribute has been changed
+function hasChanges(network, networkConfig, dataSource) {
+  let networkChanged = dataSource.network !== network
+
+  // Return directly if the network is different
+  if (networkChanged) return networkChanged
+
+  let addressChanged = (network === "near")
+    ? networkConfig.account !== dataSource.source.account
+    : networkConfig.address !== dataSource.source.address
```
otaviopace, line 125: I think there should be no chain/protocol specific logic outside of the `protocols` module/folder, and this is already there _I think_. I didn't think it through yet, what's needed to get an instance of the `Protocol` class here, but that would be preferable 🙂

src/command-helpers/network.js
dimitrovmaksim: Yeah, I guess not

src/command-helpers/network.js
```diff
@@ -0,0 +1,135 @@
...
+      let networks = {}
+
+      subgraph.dataSources.forEach(source => {
+        let sourceConfig = {[source.name]: {}}
+
+        if (source.source.address) sourceConfig[source.name]["address"] = source.source.address
+        if (source.source.startBlock) sourceConfig[source.name]["startBlock"] = source.source.startBlock
+
+        networks[source.network] = sourceConfig
+      })
```
otaviopace, line 78: This is more of a code style thing, you don't have to do it 🙂 

Personally I prefer avoiding mutability, and _I think_ this snippet does the same thing:

```javascript
const networks = subgraph.dataSources.reduce((acc, source) => ({
  ...acc,
  [source.network]: {
    [source.name]: {
      address: source.source.address,
      startBlock: source.source.startBlock,
    },
  },
}, {}))
```

Unless the `filesystem` API breaks if a field is there but the value is `underfined`.

src/command-helpers/network.js
```diff
@@ -0,0 +1,135 @@
...
+          network,
+          subgraph.dataSources,
+          networkSources,
+          networkConfig
+        )
+
+        // All data sources shoud be on the same network,
+        // so we have to update the network of all templates too.
+        if(subgraph.templates) {
+          subgraph.templates = updateTemplates(network, subgraph.templates)
```
otaviopace, line 45: Personal opinion, but I don't think you need `updateTemplates` function.

If you prefer to keep it, I would use `.map` instead of `.forEach` since you're already reasigning the value here.

Here's a way to do it, but you can do a regular `if` statement if you think it's more readable:

```javascript
subgraph.templates = subgraph.templates.map(template => ({
  ...template,
  network: template.network !== network
    ? network
    : template.network,
}))
```

In fact I don't think you even have to do the check since it will give you the same result, the Node.js engine might optimize this already. Either way I don't think this will be a performance bottleneck.

```javascript
subgraph.templates = subgraph.templates.map(template => ({
  ...template,
  network,
}))
```

src/command-helpers/network.js
```diff
@@ -0,0 +1,135 @@
...
+      await toolbox.filesystem.write(`${directory}/networks.json`, networks)
+
+      return true
+    },
+  )
+
+// Iterates through each dataSource and updates the network configuration
+// Will update only if there are changes to the configuration
+function updateSources(toolbox, spinner, network, sources, networkSources, networkConfig) {
+  sources.forEach(source => {
```
otaviopace, line 89: Same thing about `updateTemplates`, I would do `.map` instead of `.forEach` since you're already reassigning the value.

src/commands/build.js
```diff
@@ -68,13 +73,26 @@ module.exports = {
...
 
     // Show help text if requested
     if (help) {
       print.info(HELP)
       return
     }
 
+    if (network && filesystem.exists(networkFile) !== "file") {
+      print.error(`Network file '${networkFile}' does not exists!`)
+      process.exit(1)
```
otaviopace, line 89: You actually might want to do this instead:
```javascript
process.exitCode = 1
return
```

It gives the same result, but perhaps that helps better with `gluegun`? 🤔 
Not sure.

src/command-helpers/network.test.js
```diff
@@ -0,0 +1,77 @@
+const { initNetworksConfig, updateSubgraphNetwork } = require('../command-helpers/network')
```
otaviopace, line 1: Thanks for creating the tests ❤️

src/commands/build.js
dimitrovmaksim: I thought they are the same 🤔 But I guess I should stick to the current style anyways

src/command-helpers/network.js
dimitrovmaksim: I looked at it, but I wasn't sure how I could utilise it in this case

src/commands/build.js
otaviopace: They probably are.

src/command-helpers/network.js
otaviopace: So what I mean is that on the commands `build` and `init` you should pass the `protocol` instance and call the [`getContract` function](https://github.com/graphprotocol/graph-cli/blob/main/src/protocols/index.js#L191), and then the [`identifierName` one](https://github.com/graphprotocol/graph-cli/blob/main/src/protocols/ethereum/contract.js#L3), so that we don't spread `protocol` logic outside of that module.

dimitrovmaksim: Here, for the `build` command I'm checking the name of the network passed with the `--network` option.
Not sure the protocol instance will be of help in this case, but I just realised I'm not handling `near` contracts in the `init` command, and I can use it for that there.

src/command-helpers/network.js
dimitrovmaksim: I tried with `reduce`, but i think it has an issue when there's one element in the array, and it returns the dataSource instead, even with the initial value `{}`

dimitrovmaksim: I think we have to use Object.assing(), this way it works:
```javascript
const networks = subgraph.dataSources.reduce((acc, source) =>
  Object.assign(
    acc,
    {
      [source.network]: {
        [source.name]: {
          address: source.source.address,
          startBlock: source.source.startBlock,
        },
      },
    }
  )
, {})
```

src/commands/build.js
```diff
@@ -16,13 +17,15 @@ Options:
   -t, --output-format <format>  Output format for mappings (wasm, wast) (default: wasm)
       --skip-migrations         Skip subgraph migrations (default: false)
   -w, --watch                   Regenerate types when subgraph files change (default: false)
+      --network <name>          Network to use from networks.json
+      --network-file <path>    Networks file (default: "./networks.json")
```
otaviopace, line 21: I think this is missing a space before `Networks file`

otaviopace (approved):
I'll approve and merge it for now so I don't block you further, we can do the fix of the `HELP` string later.

otaviopace:
@dimitrovmaksim I just released this on `0.29.0`

Soon I'll fix the Release page in Github because we have the tags but not the release notes for them (most of them are on PRs).

PR #724, opened 2021-08-13
Migrations for specVersion 0.0.4

tilacog:
This PR adds a migration for upgrading `specVersion` 0.0.2 (and the non-canonical 0.0.3) to 0.0.4.

This new specification version enforces that all used features must be listed in the `features` key of the subgraph's manifest.

Since feature detection and validation occur inside a `graph-node` instance during deploy time, subgraphs that use any feature are expected to end with an invalid `subgraph.yml`, as `graph-cli` won't be able to infer and fill the `features` section of the manifest file automatically.

PR #860, opened 2022-03-23
Allow event handlers to require transaction receipts

tilacog:
Introduces `apiVersion` `0.0.7`, which allows event handlers to require transaction receipts to be available inside mappings,

TODO:
- [x] Migrations
- [ ] Update lockfiles as soon as `graph-ts` new version is available in external registries.
- [ ] NPM release

---
This PR is depends on
- https://github.com/graphprotocol/graph-cli/pull/724
- https://github.com/graphprotocol/graph-ts/pull/260

and is a dependency of:
- https://github.com/graphprotocol/graph-node/pull/3373

### DIFF
--- a/src/migrations/mapping_api_version_0_0_6.js
+++ b/src/migrations/mapping_api_version_0_0_6.js
@@ -21,7 +21,7 @@ module.exports = {
 
     let manifest = loadManifest(manifestFile)
     return (
-      // Only migrate if the graph-ts version is >= 0.24.0...
+      // Only migrate if the graph-ts version is >= 0.27.0...
       // Coerce needed because we may be dealing with an alpha version
       // and in the `semver` library this would not return true on equality.
       semver.gte(semver.coerce(graphTsVersion), '0.27.0') &&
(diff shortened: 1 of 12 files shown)


src/migrations/mapping_api_version_0_0_6.js
```diff
@@ -0,0 +1,74 @@
...
+      graphTsVersion = await getGraphTsVersion(sourceDir)
+    } catch (_) {
+      // If we cannot obtain the version, return a hint that the graph-ts
+      // hasn't been installed yet
+      return 'graph-ts dependency not installed yet'
+    }
+
+    let manifest = loadManifest(manifestFile)
+    return (
+      // Only migrate if the graph-ts version is >= 0.24.0...
```
evaporei, line 24: This comment and the previous migration one are incorrect.

evaporei (changes requested):

src/migrations/mapping_api_version_0_0_6.js
tilacog: Thanks! Fixed in 09f6950

evaporei (approved):
Once `graph-ts` is merged/released, you should run `yarn install` in each `examples` folder 😉

PR #872, opened 2022-04-08
codegen: generate file with contract addresses

alephao:
Hello!

I should probably have opened an issue to discuss if this is a feature you would like to have, but the implementation was so quick that I thought it would be better if I make this request with a working example.

I'll describe below why I want this feature and implementation details.

### The Problem

I have multiple subgraph manifests, one for ethereum mainnet, another for rinkeby, and another for local development. Most of the manifest contains the same configuration, but the data inside `dataSources[].source` is different. For example, on mainnet the contract address is `0x123...456`, on rinkeby it is `0x234...567` and on local it is `0x345...678`.

In my specific case, I have some conditional logic that uses the contract address to decide what to do.

https://github.com/oddworx/genzee-subgraph/blob/d40d044c5eaf4f480d66c399b59d171fbafadbc5/src/genzee.ts#L12-L27

And I currently store the contract addresses on code:

https://github.com/oddworx/genzee-subgraph/blob/d40d044c5eaf4f480d66c399b59d171fbafadbc5/src/common.ts#L6-L18

This approach is really painful, it requires me to comment/uncomment that code all the time and requires me to keep in sync addresses (if I deploy it to rinkeby again, and want to update addresses, I need to update in multiple places).

### Solution

The solution I want to propose here is for the codegen tool to generate a file `addresses.ts` that exports the addresses like so: 

```ts
export MyContractAddress = "0x123...456"
```

Projects with multiple manifest can then just run `graph codegen mymanifest.yml` and have the values updated.

### Implementation

The implementation is pretty simple, just loop through `dataSources` and create each export line using `dataSource.name` and `dataSource.source.address`.

```ts
export ${dataSource.name}Address = "${dataSource.source.address}"
```

I'm not very familiar with the codebase, but I think there might have an issue with `L60` of `src/type-generator.js` because not every source is an ABI. Instead of reading all the code and getting familiar with everything, I thought would be better if a maintainer can guide me on what we should do there.

### DIFF
--- a/src/type-generator.js
+++ b/src/type-generator.js
@@ -57,6 +57,8 @@ module.exports = class TypeGenerator {
         await this.protocolTypeGenerator.generateTypesForDataSourceTemplateABIs(templateAbis)
       }
 
+      await this.generateAddressConstants(subgraph)
+
       let schema = await this.loadSchema(subgraph)
       await this.generateTypesForSchema(schema)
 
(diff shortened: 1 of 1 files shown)


maoueh:
Templating support by network have been added recently: https://github.com/graphprotocol/graph-cli/pull/848, did you see this prior?

Is your change already handled by the PR I just linked? Is it something that is missing from the PR linked?

maoueh:
Further documentation about the issue is being worked on here also: https://github.com/graphprotocol/docs/pull/100/files

alephao:
@maoueh I think that solves my issue! I did search the issues prior but didn't saw it D:

I'll read up that PR with docs and try it out here, cheers!

alephao:
@maoueh Sorry, I rushed a bit. Re-opening this.

I checked docs and tried out the `--network` argument, but that's completely focused on deployment, which doesn't solve the issue of accessing the `contractAddress` on code.

My proposed implementation doesn't work with that `--network` argument though. I can update the code and proposal, but before I do that I'm wondering if this is something that would be welcome in the project, otherwise I'll keep as an internal tool!

azf20:
Hi @alephao ! Does the dataSource metadata ([docs](https://thegraph.com/docs/en/developer/assemblyscript-api/#data-source-metadata)) meet your needs here, or do you need to reference the address for _other_ dataSources, as well as the one which generated the event?

alephao:
@azf20 That meets my needs! Thank you!

PR #875, opened 2022-04-15
validation: Add for protocol names

evaporei:


src/validation/manifest.js
```diff
@@ -241,23 +241,61 @@ const validateValue = (value, ctx) => {
...
+// }
+const dataSourceListToMap = dataSources =>
   dataSources
-    .filter(dataSource => protocol.isValidKindName(dataSource.kind))
     .reduce(
-      (networks, dataSource) =>
-        networks.update(dataSource.network, dataSources =>
-          (dataSources || immutable.OrderedSet()).add(dataSource.name),
+      (protocolKinds, dataSource) =>
+        protocolKinds.update(dataSource.kind, networks =>
```
dimitrovmaksim, line 266: Currently if you have two datasources, one with kind ethereum and one with ethereum/contract the validation will fail. Unless this is a desired behaviour, a solution could be to replace dataSource.kind with Protocol.fromDataSources([dataSource]).name

src/validation/manifest.js
```diff
@@ -241,23 +241,61 @@ const validateValue = (value, ctx) => {
...
+          (networks || immutable.OrderedMap()).update(dataSource.network, dataSourceNames =>
+            (dataSourceNames || immutable.OrderedSet()).add(dataSource.name)),
         ),
       immutable.OrderedMap(),
     )
 
-const validateDataSourceNetworks = (value, protocol) => {
+const validateDataSourceProtocolAndNetworks = (value, protocol) => {
```
dimitrovmaksim, line 273: Do you need the `protocol` argument here, it seems unused?

evaporei: You're correct! Just removed it 🔫

src/validation/manifest.js
evaporei: Woah, good point! I guess I'll need the `protocol` parameter now 😅 

I'll probably use [Protocol.availableProtocols()](https://github.com/graphprotocol/graph-cli/blob/main/src/protocols/index.js#L25)

dimitrovmaksim (approved):

PR #878, opened 2022-04-20
feat(protocols): Arweave integration

clearloop:
## Changes

- add arweave manifest
- add arweave scaffold
- register arweave in `graph` command

## Tests

- https://github.com/chainSafe/subgraph-arweave-example

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -169,6 +169,11 @@ const processInitForm = async (
         return value
       },
     },
+    // TODO:
+    //
+    // protocols that don't support contract
+    // - arweave
+    // - tendermint
     {
       type: 'input',
       name: 'contract',
--- a/src/protocols/arweave/scaffold/manifest.js
+++ b/src/protocols/arweave/scaffold/manifest.js
@@ -1,0 +1,16 @@
+const source = ({ contract }) => `
+      account: '${contract}'`
+
+const mapping = () => `
+      apiVersion: 0.0.5
+      language: wasm/assemblyscript
+      entities:
+        - ExampleEntity
+      receiptHandlers:
+        - handler: handleReceipt
+      file: ./src/mapping.ts`
+
+module.exports = {
+  source,
+  mapping,
+}
--- a/src/protocols/arweave/scaffold/manifest.js
+++ b/src/protocols/arweave/scaffold/manifest.js
@@ -1,13 +1,15 @@
 const source = ({ contract }) => `
-      account: '${contract}'`
+      owner: '${contract}'`
 
 const mapping = () => `
       apiVersion: 0.0.5
       language: wasm/assemblyscript
       entities:
         - ExampleEntity
-      receiptHandlers:
-        - handler: handleReceipt
+      blockHandlers:
+        - handler: handleBlock
+      transactionHandlers:
+        - handler: handleTx
       file: ./src/mapping.ts`
 
 module.exports = {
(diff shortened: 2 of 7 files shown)


src/protocols/arweave/scaffold/manifest.js
```diff
@@ -0,0 +1,16 @@
+const source = ({ contract }) => `
+      account: '${contract}'`
```
evaporei, line 2: I'm not familiar with arweave terminology, do they use `account` instead of `address` just like Ethereum vs NEAR? 🤔

src/protocols/index.js
```diff
@@ -34,6 +37,7 @@ module.exports = class Protocol {
 
   static availableNetworks() {
     return immutable.fromJS({
+      arweave: ['mainnet'],
```
evaporei, line 40: The way `graph-node` currently is, I think you'll have to add a _prefix_ here, just like we've done for NEAR (`'near-mainnet'`).

Can you confirm this? @leoyvens or @maoueh

evaporei:
I'm not sure the `tests/cli/init/ethereum/from-example` submodule should be added yet, we use that folder for `init` tests, and since this isn't supported yet for `arweave` (and `tendermint`), I would leave this for another PR 🙂

src/protocols/arweave/scaffold/manifest.js
clearloop: oh it should be `owner` in arweave

clearloop:
> I'm not sure the `tests/cli/init/ethereum/from-example` submodule should be added yet, we use that folder for `init` tests, and since this isn't supported yet for `arweave` (and `tendermint`), I would leave this for another PR 🙂

tbh, I don't know how this file appears in this pr lol, I'll try to remove it

src/protocols/index.js
clearloop: `arweave-mainnet` now

src/commands/init.js
```diff
@@ -169,6 +169,11 @@ const processInitForm = async (
         return value
       },
     },
+    // TODO:
+    //
+    // protocols that don't support contract
+    // - arweave
+    // - tendermint
```
clearloop, line 176: I was trying to init an arweave subgraph example using `graph-cli` but the program crashed at this point, so I just left a `TODO` doc on this

src/protocols/arweave/scaffold/manifest.js
```diff
@@ -0,0 +1,18 @@
+const source = ({ contract }) => `
+      owner: '${contract}'`
+
+const mapping = () => `
+      apiVersion: 0.0.5
+      language: wasm/assemblyscript
+      entities:
+        - ExampleEntity
```
evaporei, line 8: Does this entity exist in the example? 🤔

clearloop: good catch!, btw, is there any file we can place our example [schema.graphql](https://github.com/ChainSafe/subgraph-arweave-example/blob/master/schema.graphql)?

evaporei: I'm not sure I'm following, can you explain with a bit more detail please?

clearloop: If we generate a `Block` entity

```yml
 entities:
        - Block
```

in the template, developers can not run the template immediately since they don't have the `schema.gql` file

evaporei: I think you're probably referring to [this](https://github.com/graphprotocol/example-subgraph/issues/34). Basically all examples should be folders in the same repo, and the `schema.graphql` with the `Block` should be there.

PR #887, opened 2022-05-09
graph add command

georg-getz:


src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+
+    // Show help text if requested
+    if (help || h) {
+      print.info(HELP)
+      return
+    }
+
+    const dataSourcesAndTemplates = await DataSourcesExtractor.fromFilePath(manifestPath)
+    let protocol = Protocol.fromDataSources(dataSourcesAndTemplates)
+    let manifest = await Subgraph.load(manifestPath, {protocol: protocol})
```
evaporei, line 70: You can simplify the `protocol: protocol` since the variable name is the same

src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+const getContractNames = (manifest) => {
+  let list = []
+  manifest.result.get('dataSources').map(dataSource => {
+    list.push(dataSource.get('name'))
+  })
+  manifest.result.get('templates').map(template => {
+    list.push(template.get('name'))
+  })
+  return list
+}
```
evaporei, line 172: I'm not the biggest fan of immutable.js, but since this project is using it, I think it would be better to either change those `.map`s to `.forEach`es since you're not using the new generated list, or they could use a `concat` + `map`:

```javascript
const getEntities = (manifest) => {
  let dataSources = manifest.result.get('dataSources')
  let templates = manifest.result.get('templates')

  return dataSources
    .concat(templates)
    .map(dataSource => dataSource.getIn(['mapping', 'entities']))
    .flatten()
}

const getContractNames = (manifest) => {
  let dataSources = manifest.result.get('dataSources')
  let templates = manifest.result.get('templates')

  return dataSources
    .concat(templates)
    .map(dataSource => dataSource.get('name'))
}
```

If for any reason there's issues with empty `List`s, `.get` receives a second argument that is the default value, which could be just a `List()`.

src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+
+    let ethabi = null
+    if (abi) {
+      ethabi = EthereumABI.load(contractName, abi)
+    } else {
+      if (network === 'poa-core') {
+        ethabi = await loadAbiFromBlockScout(EthereumABI, network, address)
+      } else {
+        ethabi = await loadAbiFromEtherscan(EthereumABI, network, address)
+      }
```
evaporei, line 92: Not scope of this PR, but just noting that we should probably centralize this logic somewhere, since I think I've seen this `poa-core` if more than a couple of times 😆

src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    let updateResult = updateEventNamesOnCollision(ethabi, entities, contractName, mergeEntities)
+    let collisionEntities = updateResult.collisionEntities
+    let onlyCollisions = updateResult.onlyCollisions
+    ethabi.data = updateResult.abiData
```
evaporei, line 98: You could do a destructuring here :)

```suggestion
    let { collisionEntities, onlyCollisions, abiData } = updateEventNamesOnCollision(ethabi, entities, contractName, mergeEntities)
    ethabi.data = abiData
```

src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+      let mapping = firstDataSource.get('mapping').asMutable()
+
+      // Save the address of the new data source
+      source.abi = firstDataSource.get('source').get('abi')
+
+      dataSource.set('mapping', mapping)
+      dataSource.set('source', source)
+    }
+
+    result.set('dataSources', dataSources.push(dataSource))
```
evaporei, line 121: I'm a bit confused here, is this `.push` from JS? Does it return the whole list mutated? 🤔

src/commands/add.js
```diff
@@ -0,0 +1,208 @@
...
+
+    await Subgraph.write(result, manifestPath)
+    manifest = await Subgraph.load(manifestPath, {protocol: protocol})
```
evaporei, line 124: Same as the other simplification, `{protocol: protocol}` -> `{ protocol }`

src/command-helpers/scaffold.js
```diff
@@ -59,8 +79,54 @@ const writeScaffold = async (scaffold, directory, spinner) => {
   await writeScaffoldDirectory(scaffold, directory, spinner)
 }
 
+const writeABI = async (abi, contractName, abiPath) => {
+  let data = prettier.format(JSON.stringify(abi.data), {
+    parser: 'json',
+  })
+  let filePath = abiPath ? abiPath : `./abis/${contractName}.json`
```
evaporei, line 86: Maybe this could go into the function signature (or just turn the ternary into an `||`)? 🤔 

```javascript
const writeABI = async (abi, contractName, abiPath = `./abis/${contractName}.json`) => {
```

src/command-helpers/scaffold.js
```diff
@@ -59,8 +79,54 @@ const writeScaffold = async (scaffold, directory, spinner) => {
...
+
+  await fs.writeFile(filePath, data, { encoding: 'utf-8' })
+}
+
+const writeSchema = async (abi, protocol, schemaPath, entities) => {
+  const events = protocol.hasEvents()
+    ? abiEvents(abi).filter(event => entities.indexOf(event.get('name')) === -1).toJS()
```
evaporei, line 93: The reason you do a `.toJS()` here is something to due with the `prettier` API?

src/command-helpers/scaffold.js
```diff
@@ -59,8 +79,54 @@ const writeScaffold = async (scaffold, directory, spinner) => {
...
+      parser: 'graphql',
+    },
+  )
+
+  await fs.appendFile(schemaPath, data, { encoding: 'utf-8' })
+}
+
+const writeMapping = async (abi, protocol, contractName, entities) => {
+  const events = protocol.hasEvents()
+    ? abiEvents(abi).filter(event => entities.indexOf(event.get('name')) === -1).toJS()
```
evaporei, line 110: Same question as the other `.toJS()` one above

evaporei (commented):
One last question related to the ["spec"](https://hackmd.io/Fc0gQXHDT8-OtqVf76e23w?view) is, will we include the behavior below in `graph init`?

スクリーンショット 2022-05-12 10 40 40

src/commands/add.js
georg-getz: It's not, it's from Immutable.js List and yes, it returns the mutated list.

src/command-helpers/scaffold.js
georg-getz: No, it's because `abi.data` is an immutable.js type and if I don't turn it into a JS type the `generateEventType` function crashes:

image

src/commands/add.js
georg-getz: I added the empty list arg as the default return type is `undefined`

src/command-helpers/abi.js
```diff
@@ -0,0 +1,69 @@
...
+      // case of success
+      if (json.status === '1') {
+        return new ABI('Contract', undefined, immutable.fromJS(JSON.parse(json.result)))
+      } else {
+        throw new Error('ABI not found, try loading it from a local file')
+      }
+    },
+  )
+
+const getEtherscanLikeAPIUrl = (network) => {
```
evaporei, line 52: Thanks 😊

PR #888, opened 2022-05-09
Rename Tendermint to Cosmos

pienkowb:


src/protocols/cosmos/scaffold/manifest.js
```diff
@@ -0,0 +1,16 @@
+const source = () => `
+      startBlock: 0`
```
evaporei, line 2: Isn't `startBlock` zero by default? 🤔

pienkowb: So you're suggesting to remove `startBlock` entirely (as it's `0` by default)?

evaporei: Yes!

evaporei (approved):

evaporei:
@pienkowb it seems you'll need to solve conflicts with the `main` branch.

PR #897, opened 2022-05-27
Cache matchstick latest version

dimitrovmaksim:
closes https://github.com/LimeChain/matchstick/issues/328

Some users mentioned that every time `graph test` is executed a request to `https://api.github.com/repos/LimeChain/matchstick/releases/latest` is made to fetch the latest matchstick version tag, which is an overkill. This PR adds something like a in-file cache, where the latest version tag and a timestamp are saved to `<tests_folder>/.latest.json` 
Now this request will be executed only if no version is passed with `-v/--version` option and no cache file is found or the cached version is older than 1 day. 

This PR also adds support for linux v22 binaries (will be available when matchstick v0.5.0 is released in the coming days)
This PR also includes slight code refactoring


evaporei (approved):
LGTM

fredlacs:
can we declare this in the `subgraph.yaml` instead? would be easier to manage

dimitrovmaksim:
> can we declare this in the `subgraph.yaml` instead? would be easier to manage

I don't think `subgraph.yaml` is the appropriate place for this, as it doesn't relate to the subgraph itself

fredlacs:
is there a different config file for the `graph-cli` I could use that sets this so I don't need to pass in the variable every time?

dimitrovmaksim:
If you mean to set the matchstick version in a config, currently there's not such option. There is a [matchstick config](https://thegraph.com/docs/en/developer/matchstick/#configuration), but it doesn't allow to set the version there. What you can do is add a script in your `package.json`, for example: `"test": "graph test -v <version>"` and run it with `yarn test` or `npm run test`

PR #904, opened 2022-06-07
Add additional contracts on graph init

ge0rgiev:
```Changes to the existing graph init flow:

At the end of adding --from-contract, ask if the developer wants to add another contract (and if so, initiate graph add for that contract)
```

as per the spec https://hackmd.io/Fc0gQXHDT8-OtqVf76e23w?view

### DIFF
--- a/src/commands/init.js
+++ b/src/commands/init.js
@@ -854,7 +854,6 @@ const addAnotherContract = async (toolbox, { protocolInstance, directory }) => {
         type: 'input',
         name: 'abi',
         message: 'ABI file (path)',
-        initial: abi,
         skip: () => abiFromFile === false
       },
       {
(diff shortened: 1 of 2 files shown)


src/commands/init.js
```diff
@@ -815,5 +814,84 @@ const initSubgraphFromContract = async (
...
+        if (abi.includes(directory)) {
+          commandLine.push('--abi', path.normalize(abi.replace(directory, '')))
+        } else {
+          commandLine.push('--abi', abi)
+        }
+      }
+  
+      await graphCli.run(commandLine)
+  
+      process.chdir(cwd)
```
evaporei, line 889: Perhaps this should be in the `finally` statement so it will happen no matter if the other lines fail or not.

evaporei (approved):

src/commands/init.js
evaporei: With this change, I think you've forgot to move the `const cwd =` above the `try` block.

PR #911, opened 2022-06-17
add partial account filters for near

mangas:


tests/cli/validation/near-is-valid/subgraph.yaml
```diff
@@ -10,6 +10,14 @@ dataSources:
     source:
       account: wnear.flux-dev
       startBlock: 1
+      accounts:
```
evaporei, line 13: Can `account` and `accounts` co-exist on the same manifest?

mangas: yep it's a logical OR

evaporei: Just to confirm I got it right, so on `NEAR` subgraphs we won't have a single "address" (account) per data source like in Ethereum/EVM-compat ones right?

evaporei (approved):

PR #919, opened 2022-06-29
Generate a tests helper file on init or add

dimitrovmaksim:
Generate tests helpers files containing functions that will create a mocked Event for each ABI. For example initializing a subgraph for the Gravatar contract from the example-subgraph will generate a helper file `/tests/gravity-utils.ts` containing the following functions:

```
import { newMockEvent } from "matchstick-as"
import { ethereum, BigInt, Address } from "@graphprotocol/graph-ts"
import { NewGravatar, UpdatedGravatar } from "../generated/Gravity/Gravity"

export function createNewGravatarEvent(
  id: BigInt,
  owner: Address,
  displayName: string,
  imageUrl: string
): NewGravatar {
  let newGravatarEvent = changetype<NewGravatar>(newMockEvent())

  newGravatarEvent.parameters = new Array()

  newGravatarEvent.parameters.push(
    new ethereum.EventParam("id", ethereum.Value.fromUnsignedBigInt(id))
  )
  newGravatarEvent.parameters.push(
    new ethereum.EventParam("owner", ethereum.Value.fromAddress(owner))
  )
  newGravatarEvent.parameters.push(
    new ethereum.EventParam(
      "displayName",
      ethereum.Value.fromString(displayName)
    )
  )
  newGravatarEvent.parameters.push(
    new ethereum.EventParam("imageUrl", ethereum.Value.fromString(imageUrl))
  )

  return newGravatarEvent
}

export function createUpdatedGravatarEvent(
  id: BigInt,
  owner: Address,
  displayName: string,
  imageUrl: string
): UpdatedGravatar {
  let updatedGravatarEvent = changetype<UpdatedGravatar>(newMockEvent())

  updatedGravatarEvent.parameters = new Array()

  updatedGravatarEvent.parameters.push(
    new ethereum.EventParam("id", ethereum.Value.fromUnsignedBigInt(id))
  )
  updatedGravatarEvent.parameters.push(
    new ethereum.EventParam("owner", ethereum.Value.fromAddress(owner))
  )
  updatedGravatarEvent.parameters.push(
    new ethereum.EventParam(
      "displayName",
      ethereum.Value.fromString(displayName)
    )
  )
  updatedGravatarEvent.parameters.push(
    new ethereum.EventParam("imageUrl", ethereum.Value.fromString(imageUrl))
  )

  return updatedGravatarEvent
}

```

### DIFF
--- a/src/command-helpers/scaffold.js
+++ b/src/command-helpers/scaffold.js
@@ -121,7 +121,7 @@ const writeMapping = async (abi, protocol, contractName, entities) => {
   await fs.writeFile(`./src/${strings.kebabCase(contractName)}.ts`, mapping, { encoding: 'utf-8' })
 }
 
-const writeTestsHelper = async (abi, contractName, directory = "./") => {
+const writeTestsHelper = async (abi, contractName, directory = "") => {
   let utilsFile = prettier.format(
     generateTestHelperFile(abiEvents(abi).toJS(), contractName),
     { parser: 'typescript', semi: false },
--- a/src/command-helpers/scaffold.js
+++ b/src/command-helpers/scaffold.js
@@ -7,6 +7,7 @@ const { step } = require('./spinner')
 const Scaffold = require('../scaffold')
 const { generateEventIndexingHandlers } = require('../scaffold/mapping')
 const { generateEventType, abiEvents } = require('../scaffold/schema')
+const { ascTypeForEthereum, ethereumFromAsc } = require("../codegen/types")
 const { strings } = require('gluegun')
 const { Map } = require('immutable')
 
@@ -120,6 +121,63 @@ const writeMapping = async (abi, protocol, contractName, entities) => {
   await fs.writeFile(`./src/${strings.kebabCase(contractName)}.ts`, mapping, { encoding: 'utf-8' })
 }
 
+const writeTestsHelper = async (abi, contractName, directory = "./") => {
+  let utilsFile = prettier.format(
+    generateTestHelperFile(abiEvents(abi).toJS(), contractName),
+    { parser: 'typescript', semi: false },
+  )
+
+  const filePath = path.join(directory, `tests/${strings.kebabCase(contractName)}-utils.ts`)
+  await fs.writeFile(filePath, utilsFile, { encoding: 'utf-8' })
+}
+
+const generateTestHelperFile = (events, contractName) => {
+  const eventsNames = events.map(event => event.name)
+  const eventsTypes = events.flatMap(event => event.inputs.map(input => ascTypeForEthereum(input.type))).filter(type => !isNativeType(type))
+  const importedTypes = [...new Set(eventsTypes)].join(', ');
+
+  let utils = `import { newMockEvent } from 'matchstick-as';
+  import { ethereum, ${importedTypes} } from '@graphprotocol/graph-ts';
+  import { ${eventsNames.join(', ')} } from '../generated/${contractName}/${contractName}';
+  `
+
+  events.forEach(function(event) {
+    utils = utils.concat("\n", generateMockedEvent(event))
+  });
+
+  return utils
+}
+
+const generateMockedEvent = (event) => {
+  const varName = `${strings.camelCase(event.name)}Event`
+  const fnArgs = event.inputs.map(input => `${input.name}: ${ascTypeForEthereum(input.type)}`);
+  const ascToEth = event.inputs.map(input => `${varName}.parameters.push(new ethereum.EventParam("${input.name}", ${ethereumFromAsc(input.name, input.type)}))`);
+
+  return  `
+    export function create${event._alias}Event(${fnArgs.join(', ')}): ${event._alias} {
+      let ${varName} = changetype<${event._alias}>(newMockEvent());
+
+      ${varName}.parameters = new Array();
+
+      ${ascToEth.join('\n')}
+
+      return ${varName};
+    }
+  `
+
+}
+
+const isNativeType = (type) => {
+  let natives = [
+    /Array<([a-zA-Z0-9]+)?>/,
+    /i32/,
+    /string/,
+    /boolean/
+  ]
+
+  return natives.some(rx => rx.test(type));
+}
+
 module.exports = {
   ...module.exports,
   generateScaffold,
@@ -128,4 +186,5 @@ module.exports = {
   writeABI,
   writeSchema,
   writeMapping,
+  writeTestsHelper,
 }
--- a/src/scaffold/index.js
+++ b/src/scaffold/index.js
@@ -137,6 +137,12 @@ dataSources:
       : undefined
   }
 
+  initTestsFolder() {
+    return this.protocol.hasEvents()
+    ? {}
+    : undefined
+  }
+
   generate() {
     return {
       'package.json': this.generatePackageJson(),
@@ -145,7 +151,7 @@ dataSources:
       'tsconfig.json': this.generateTsConfig(),
       src: { [`${strings.kebabCase(this.contractName)}.ts`]: this.generateMapping() },
       abis: this.generateABIs(),
-      tests: {},
+      tests: this.initTestsFolder(),
     }
   }
 }
--- a/src/scaffold/index.js
+++ b/src/scaffold/index.js
@@ -145,6 +145,7 @@ dataSources:
       'tsconfig.json': this.generateTsConfig(),
       src: { [`${strings.kebabCase(this.contractName)}.ts`]: this.generateMapping() },
       abis: this.generateABIs(),
+      tests: {},
     }
   }
 }
--- a/src/commands/add.js
+++ b/src/commands/add.js
@@ -5,7 +5,7 @@ const { withSpinner } = require('../command-helpers/spinner')
 const Subgraph = require('../subgraph')
 const Protocol = require('../protocols')
 const DataSourcesExtractor = require('../command-helpers/data-sources')
-const { generateDataSource, writeABI, writeSchema, writeMapping } = require('../command-helpers/scaffold')
+const { generateDataSource, writeABI, writeSchema, writeMapping, writeTestsHelper } = require('../command-helpers/scaffold')
 const { loadAbiFromEtherscan, loadAbiFromBlockScout } = require('../command-helpers/abi')
 const EthereumABI = require('../protocols/ethereum/abi')
 const { fixParameters } = require('../command-helpers/gluegun')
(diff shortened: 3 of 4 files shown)


src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
...
+  "BigInt": 234,
+  "Bytes": 1234567890,
+  "Address": "0x0000000000000000000000000000000000000001",
+  "string": "Example string value",
+  "bool": true,
+}
+
+const generateTestsFiles = (contract, events, indexEvents) => {
+  const event = events[0]
+  const eventsTypes = events.flatMap(event => event.inputs.map(input => ascTypeForEthereum(input.type))).filter(type => !type.startsWith("ethereum.") && !isNativeType(type))
```
evaporei, line 16: I appreciate that most files your text editor configuration is fixing/improving the formatting of the code 😊 
Though for this specific line it's a bit hard to read, can you split it up please? It could be something similar to this example:

```javascript
const eventsTypes = events
  .flatMap(event =>
    event
      .inputs
      .map(input => ascTypeForEthereum(input.type))
  ).filter(type => !type.startsWith("ethereum.") && !isNativeType(type))
```

src/scaffold/index.js
```diff
@@ -48,10 +49,12 @@ module.exports = class Scaffold {
             `--node http://localhost:8020/ ` +
             `--ipfs http://localhost:5001 ` +
             this.subgraphName,
+          'test': 'graph test',
         },
         dependencies: {
           '@graphprotocol/graph-cli': GRAPH_CLI_VERSION,
           '@graphprotocol/graph-ts': `0.27.0`,
+          'matchstick-as': `0.5.0`
```
evaporei, line 57: Nice! \o/ 🎉

src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
...
+  "i32": 123,
+  "BigInt": 234,
+  "Bytes": 1234567890,
+  "Address": "0x0000000000000000000000000000000000000001",
+  "string": "Example string value",
+  "bool": true,
+}
+
+const generateTestsFiles = (contract, events, indexEvents) => {
+  const event = events[0]
```
evaporei, line 15: Why is this variable created? 🤔

src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
...
+    // https://thegraph.com/docs/en/developer/matchstick/#write-a-unit-test
+
+    test("${entity} created and stored", () => {
+      assert.entityCount('${entity}', 1)
+
+      // 0xa16081f360e3847006db660bae1c6d1b2e17ec2a is the default address used in newMockEvent() function
+      ${generateFieldsAssertions(entity, eventInputs, indexEvents)}
+
+    // More assert options:
+    // https://thegraph.com/docs/en/developer/matchstick/#asserts
```
evaporei, line 140: It _seems_ like these two line comments are missing a TAB.

src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
...
+  import { assert, describe, test, clearStore, beforeAll, afterAll} from "matchstick-as/assembly/index"
+  import { ${importTypes} } from "@graphprotocol/graph-ts"
+  import { ${entity} } from "../generated/schema"
+  import { ${indexEvents ? `${eventName} as ${eventName}Event` : eventName} } from "../generated/${contract}/${contract}"
+  import { handle${eventName} } from "../src/${strings.kebabCase(contract)}"
+  import { create${eventName}Event } from "./${strings.kebabCase(contract)}-utils"
+
+
+  // Tests structure (matchstick-as >=0.5.0)
+  // https://thegraph.com/docs/en/developer/matchstick/#tests-structure-0-5-0
```
evaporei, line 117: I thought about maybe in the future `matchstick` providing the test file as a template (eg: `handlebars`), but maybe it's better to generate it fully here in the `graph-cli`. Just wondering really, for now I think the way it is at the moment is great 🙂 👍

src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
+const prettier = require('prettier')
+const { strings } = require('gluegun')
```
evaporei, line 2: I didn't realize `gluegun` had those helpers, pretty awesome!

src/scaffold/tests.js
dimitrovmaksim: Yeah, you're right it is unneeded. I merged some code from another PR and reworked some things and I guess I missed that. Thanks for pointing it out!

src/scaffold/tests.js
```diff
@@ -0,0 +1,185 @@
...
+      innerType = type.match(/Array<(.*?)>/)[1]
+      return `[${expectedValue(innerType)}]`
+    default:
+      let value = VARIABLES_VALUES[type]
+      return value ? value : `${type} Not implemented`
+  }
+}
+
+// Checks if the type is a native AS type or should be imported from graph-ts
+const isNativeType = type => {
```
evaporei, line 90: Should this function be considering all the types described [here](https://github.com/graphprotocol/graph-cli/blob/main/src/codegen/typescript.js#L106)? Maybe there's an overlap on what both do.

evaporei (commented):
I would just like to comment I really like the code style here overall 👌 Though that's a bit of personal preference, JavaScript has so many was to do the same thing 😅

src/scaffold/tests.js
dimitrovmaksim: We thought that it is more fitting to add it to graph-cli, because it handles generation of other similar files and already has the helpers needed (the conversion functions provided by `src/codegen/types` were a massive help)

src/scaffold/tests.js
dimitrovmaksim: Here i'm checking against the types after ethereum.Value to AS conversions https://github.com/graphprotocol/graph-cli/blob/main/src/codegen/types/conversions.js#L6 and i have to filter out the ones that are AS native and leave only the ones that get imported from graph-ts. In this case the only native AS types are boolean, Array<T>, i32 and string.

dimitrovmaksim:
> I would just like to comment I really like the code style here overall 👌 Though that's a bit of personal preference, JavaScript has so many was to do the same thing 😅

I have a good teacher 🙇🏻 😄

src/scaffold/tests.js
dimitrovmaksim: But actually I just realised something i've missed, I don't handle arrays of graph-ts types, currently it will filter them. For example `Array<BigInt>` wi'll have to check the inner type too.

src/scaffold/tests.js
dimitrovmaksim: This variable was created because the example test is created only for the first event, similarly how in the mappings only the first event has a full handler example. But it can be removed anyways and just pass events[0] to `generateExampleTest()` function

src/scaffold/index.js
KholdStare: Shouldn't `matchstic-as` and `@graphprotocol/graph-cli` be `dev-dependencies`?

src/scaffold/ethereum.test.js
```diff
@@ -243,6 +243,265 @@ export function handleExampleEvent1(event: ExampleEvent1Event): void {
...
+  })
+
+  // For more test scenarios, see:
+  // https://thegraph.com/docs/en/developer/matchstick/#write-a-unit-test
+
+  test(\"ExampleEvent created and stored\", () => {
+    assert.entityCount(\"ExampleEvent\", 1)
+
+    // 0xa16081f360e3847006db660bae1c6d1b2e17ec2a is the default address used in newMockEvent() function
+    assert.fieldEquals(
```
KholdStare, line 423: Maybe a dumb question - but why use the "stringly typed" `fieldEquals`, as opposed to doing

```typescript
const event = ExampleEvent.load("0xa16081f360e3847006db660bae1c6d1b2e17ec2a-1");
assert.stringEquals("Example string value", event.param2);

// etc...
```

src/command-helpers/scaffold.js
```diff
@@ -106,18 +113,33 @@ const writeSchema = async (abi, protocol, schemaPath, entities) => {
...
+    parser: 'typescript',
+    semi: false,
+  })
+
+  await fs.writeFile(`./src/${strings.kebabCase(contractName)}.ts`, mapping, {
+    encoding: 'utf-8',
+  })
+}
+
+const writeTestsFiles = async (abi, contractName) => {
```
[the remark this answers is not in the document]
dimitrovmaksim: It is not part of codegen

src/scaffold/ethereum.test.js
dimitrovmaksim: You don't need to load the entities with `fieldEquals`. Also after loading the entity you'll have to make sure that it is not `null` before trying to access the `param2` field, while fieldEquals also handles that case.

With your example you'll have to do this:

```
const event = ExampleEvent.load("0xa16081f360e3847006db660bae1c6d1b2e17ec2a-1");
if(event) {
  assert.stringEquals("Example string value", event.param2);
}
```

src/command-helpers/scaffold.js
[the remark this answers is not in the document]
dimitrovmaksim: Also the utils files are not necessarily static, because the mocked events have default values, which someone may want to change, for example the `address`

src/scaffold/ethereum.test.js
KholdStare: True, but then I don't need to retype `"0xa16081f360e3847006db660bae1c6d1b2e17ec2a-1"` and `"ExampleEvent"` over and over again. You also get autocompletion and typechecking on the parameters. 

If you misspell `"param2"` as `"parm2"` you'll get a test failure as opposed to a compilation error.

```
    assert.fieldEquals(
      "ExampleEvent",
      "0xa16081f360e3847006db660bae1c6d1b2e17ec2a-1",
      "parm2",
      "Example string value"
    )
```

src/command-helpers/scaffold.js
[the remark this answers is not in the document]
KholdStare: Ok. If it's before codegen, it's still referencing files from the codegen step (e.g. `import { ExampleEvent } from "../generated/schema"`) . Seems like you would want to regenerate the utils file every time `codegen` is run no?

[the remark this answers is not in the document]
dimitrovmaksim: This is the entity the `init` command will generate for you. In the test case you are referencing, the entity is called `ExampleEvent`, because it tests when the `indexEvents` option is set to `true`. In this case `init` will generate entities for each event with the same name as the event. And to answer and your other question, those are example files with example tests that are meant to match the mapping generated by init. You'll probably have to update them yourself to match your logic. Those are meant to lay a foundation and get you started.

src/scaffold/tests.js
evaporei: Maybe this code generation should improve on the https://github.com/graphprotocol/graph-cli/blob/main/src/codegen/typescript.js code we have, so that it's easier to navigate and make assumptions about types.

But that can be improved in a future PR 🙂

evaporei (approved):

PR #938, opened 2022-07-24
Ensure contract actually has events when generating example tests

dimitrovmaksim:
Currently the example test generation will break if `abiEvents(abi).toJS()` returns an empty array. This PR fixes that case.

Side note, `codegen` will also break if the contract abi has no events. The manifest scaffold loops through the list of events returned by `abiEvents(abi).toJS()`. If it's an empty array the `entities` and `eventHandlers` keys will have empty value, which will break `codegen`. The `entities` can be fixed with adding a check and setting `ExampleEntity` if the events array is empty. For the `eventHandlers` the case is not that easy, because you can't set a random event there, as `codegen` will check if that event exists in the contract ABI and will fail if not. If the whole `eventHandlers` key is omitted ` codegen` will again fail because it requires one of eventHandlers, callHandlers or blockHandlers to be declared in the manifest file. Maybe the best solution in this case will be to skip `codegen` at all.

### DIFF
--- a/src/command-helpers/scaffold.js
+++ b/src/command-helpers/scaffold.js
@@ -128,8 +128,11 @@ const writeMapping = async (abi, protocol, contractName, entities) => {
   })
 }
 
-const writeTestsFiles = async (abi, contractName) => {
-  const events = abiEvents(abi).toJS()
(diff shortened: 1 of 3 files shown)


evaporei (approved):
LGTM!

PR #942, opened 2022-08-08
feat: support M2

0xthk:


maoueh (approved):
I imagine at some point the binary name `binary-macos-11-m1` should change also, I'm really clueless however about where it's coming from. Not a big deal anyway for this PR.

PR #955, opened 2022-09-06
Add support for multiple subgraph examples

### LINKED ISSUE #940: `graph init` to support more varied examples
`graph init --from-example path/to/example`

- Migrate to use `graphprotocol/example-subgraphs` instead of `graphprotocol/example-subgraph`
- The path should be relative to the base of the repository
- Default to the Ethereum Gravatar example if no path is provided

pienkowb:
Resolves #940


evaporei (approved):
LGTM

PR #960, opened 2022-09-14
Update studio.js to allow Gnosis deployments

azf20:


evaporei:
@azf20 for some reason the build is failing for Node.js 12 and 14, even though the project is working normally in those versions

スクリーンショット 2022-09-21 10 59 30


We can merge it while I don't fix it

PR #1004, opened 2022-11-29
migrate to GH Actions

saihaj:
Upgraded `assemblyscript` to `0.19.23` to fix https://github.com/AssemblyScript/assemblyscript/issues/1948 which was causing tests to fail. Node 14 seems to take way longer. We do need to investigate tests and improve them because there are still improvements. But that is a future thing for us to tackle. For now we should have way less flaky CI.


evaporei (commented):
Everything looks alright here too :blush: congrats on the work :clap: 
Before merging this, we should double check if:

- It's ok to upgrade AssemblyScript. I tried upgrading to a patch version similar to yours in the past and had issues when indexing a subgraph.
- We can upgrade the docker image to `latest` and remove those version environment variables :pray:

PR #1014, opened 2022-12-06
chore: setup changeset for release

saihaj:
Changeset is what we @the-guild-org use to manage all our package releases. This helps us release multiple manages from monorepo and this tool take care of updating version for any dependent packages. We have a shared config which we utilize just to simplify the workflow. You can read more here https://github.com/the-guild-org/shared-config#shared-config-by-the-guild

closes https://github.com/graphprotocol/graph-cli/issues/1009


.github/workflows/pr.yml
```diff
@@ -0,0 +1,21 @@
...
+    uses: the-guild-org/shared-config/.github/workflows/changesets-dependencies.yaml@main
+    secrets:
+      githubToken: ${{ secrets.GITHUB_TOKEN }}
+
+  release:
+    uses: the-guild-org/shared-config/.github/workflows/release-snapshot.yml@main
+    with:
+      npmTag: alpha
+      buildScript: build
+      nodeVersion: 18
```
dotansimha, line 18: Does this mean that CI and this repo work fine with Node 18? :)

dotansimha (approved):

package.json
```diff
@@ -2,6 +2,8 @@
   "name": "graphprotocol-tools-monorepo",
   "private": true,
   "devDependencies": {
+    "@changesets/changelog-github": "^0.4.7",
+    "@changesets/cli": "^2.25.2",
```
evaporei, line 6: Maybe being picky here, but what do you guys think of locking versions? We had an issue in the past with security related to a dependency and then we decided to start to pin them.

evaporei (approved):

package.json
saihaj: these are dev dependencies and we keep `renovate` bot to keep them up to date

PR #1020, opened 2022-12-13
refactor: move to typescript, drop eslint and use prettier for linting, fix linting issues

saihaj:
- Moved all files to TypeScript, most types are `any` to expedite the process
- Removed eslint because its only rule is prettier
- Lint with prettier instead
- Fixed all linting issues


saihaj:
LGTM

enisdenjo (approved):

packages/cli/src/codegen/types/index.ts
```diff
@@ -4,17 +4,21 @@ import TYPE_CONVERSIONS from './conversions'
 
 // Conversion utilities
 
-const conversionsForTypeSystems = (fromTypeSystem, toTypeSystem) => {
+const conversionsForTypeSystems = (fromTypeSystem: string, toTypeSystem: string) => {
   let conversions = TYPE_CONVERSIONS.getIn([fromTypeSystem, toTypeSystem])
-  if (conversions === undefined) {
+  if (conversions == undefined) {
```
saihaj, line 9: maybe we should keep this one? maybe it was trying to account for all `falsy` values?

saihaj (commented):
LGTM

packages/cli/src/codegen/types/index.ts
enisdenjo: It was strictly checking for undefined previously. I've reverted the change just to reduce the changeset (and keep the bugs away).

See here: https://github.com/graphprotocol/graph-cli/pull/1020/commits/954cee6d1fdda14e027b8eeeea38fa09b255b23c.

PR #1021, opened 2022-12-14
feat: remove immutable

saihaj:


enisdenjo:
I'd say we keep ImmutableJS for now. This turned out to be a much bigger refactor than I expected, a lot of internals have ImmutableJS-like behaviour - some of them originating from `graph-ts`.

I worry that the CLI might become unstable with the size of this migration, we should queue up a breaking release. @saihaj @dotansimha thoughts?

dotansimha:
> I worry that the CLI might become unstable with the size of this migration, we should queue up a breaking release. @saihaj @dotansimha thoughts?

I think we can keep it for now, and just see if we can make it work fine with the TS migration. What breaking release you had in mind?

enisdenjo:
> What breaking release you had in mind?

Due to the size of the refactor, I have a feeling that we shouldn't push it as a minor or patch even if the user-facing API does not change.

PR #1037, opened 2023-01-17
chore: bring in graph-ts

saihaj:
Bringing in https://github.com/graphprotocol/graph-ts to this monorepo to unify tooling.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,6 +20,7 @@
     "lint:prettier": "prettier -c .",
     "release": "pnpm build && changeset publish",
     "test:cli": "pnpm --filter @graphprotocol/graph-cli test",
+    "test:ts": "pnpm --filter @graphprotocol/graph-ts test",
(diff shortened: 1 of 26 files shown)


package.json
```diff
@@ -14,7 +14,7 @@
     "pnpm": ">=7.14.2"
   },
   "scripts": {
-    "build": "pnpm --filter=@graphprotocol/graph-cli build",
+    "build": "pnpm --filter=@graphprotocol/graph-* build",
```
dotansimha, line 17: @saihaj I guess we can just do without filter? this way it runs on all monorepos packages that has `build` script?  (@enisdenjo the pnpm master, to confirm 😆 )

saihaj: I did try that but issue is the examples and tests are generated which would error since they do not have `graph-cli` built since all this is done concurrently

evaporei (approved):

dotansimha (approved):

PR #1040, opened 2023-01-18
fix: target Node 14 and specify engine in package.json

enisdenjo:
### TODO

- [x] Is this a breaking change? Do we want to support even lower Node versions? See https://github.com/graphprotocol/graph-cli/pull/1040#pullrequestreview-1254212875.


saihaj (approved):
supporting LTS version I don't think we need to go older than this

PR #1059, opened 2023-02-05
Auto fetch startBlock

incrypto32:
Auto fetch start block
- Use etherscan API to fetch contract deploy transaction 
- Get the blockNumber using `eth_getTransactionByHash` by calling a public RPC endpoint


packages/cli/src/command-helpers/constants.ts
```diff
@@ -0,0 +1,29 @@
+export const PUBLIC_RPC_ENDPOINTS: any = {
```
schmidsi, line 1: Although I like this pattern better than what we have in [`getEtherscanLikeAPIUrl`](https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/command-helpers/abi.ts#L58), I think consistency is key. So I suggest to either refactor `getEtherscanLikeAPIUrl` to use constants too.

schmidsi:
This is great! Only left some small comments. What happens if these requests time-out?

packages/cli/src/command-helpers/constants.ts
saihaj: agreed I think what we have`getEtherscanLikeAPIUrl` is more safe

.vscode/settings.json
```diff
@@ -1,5 +0,0 @@
-{
```
saihaj, line 1: I don't think we need to remove this

saihaj (commented):
if you push a `changeset` it will automatically create alpha releases which you can experiment on some app

packages/cli/src/commands/init.ts
```diff
@@ -268,6 +269,15 @@ const processInitForm = async (
...
         }
+        // If startBlock is not set, try to load it.
+        if (!startBlock) {
+          try {
+            // Load startBlock for this contract
+            startBlock = Number(await loadStartBlockForContract(network!, value)).toString();
+          } catch (error) {
+            // noop
+          }
+        }
```
saihaj, line 280: we don't follow a `retryAsync` pattern anywhere but ideally we put this logic in a retry logic and that would also mean we reflect in the prompt UI. A future improvement in CLI.

<details>
<summary>retryAsync</summary>

```ts
async function retryAsync(fn, {
  tries,
  interval,
  onError
}) {
  let currentTry = 0;
  let latestError;
  while (currentTry < tries) {
    try {
      return await fn({
        attempt: currentTry + 1
      });
    } catch (error) {
      onError == null ? void 0 : onError({ error, currentTry });
      latestError = error;
      currentTry += 1;
      if (currentTry !== tries) {
        await wait(interval);
      }
    }
  }
  throw latestError || new Error(`Invalid retries: ${tries}`);
}
```
<details>

saihaj: https://github.com/graphprotocol/graph-tooling/issues/1060

.vscode/settings.json
incrypto32: Oops, agreed. I just Re added it

packages/cli/src/command-helpers/constants.ts
incrypto32: @saihaj @schmidsi how do you propose going about this? Merge this as it is for now? And maybe migrate to a cleaner way in a future PR?

There are other places that could use some consistency, currently the list of available network is sort of hardcoded 
[here](https://github.com/graphprotocol/graph-tooling/blob/74bb29c7125930c7d0817861340ab9bed349517b/packages/cli/src/protocols/index.ts#L76).

If we could make this a single array with objects that has the etherscan URL and the public RPC endpoint all combined into a single object that is used everywhere where we need the list of supported networks, their etherscan/public RPC endpoint, that would be good in my opinion. Let me know what you folks think

saihaj: I think for this one let's just copy `getEtherScanLikeAPIUrl` way cause it is better than doing `any` type here. I do agree we need to clean this all up and have networks as a constant that we can use anywhere.

packages/cli/src/commands/init.ts
incrypto32: Agree 100%, would love to work on it, i will write in the issue, would need your guidance on how to go about it.

packages/cli/src/command-helpers/abi.test.ts
```diff
@@ -0,0 +1,99 @@
...
+  moonriver: {
+    '0x3dB01570D97631f69bbb0ba39796865456Cf89A5': 800_950,
+  },
+  optimism: {
+    '0xc35DADB65012eC5796536bD9864eD8773aBc74C4': 7_019_815,
+  },
+
+  // Skipping these networks for now because they do not support the latest etherscan contracts API or is blockScout based
+
+  // Chains that doesnt support the latest etherscan contracts API
```
saihaj, line 41: is it worth putting a `warning` for a user who are going to use these networks?

saihaj (approved):

saihaj:
Thanks @incrypto32! and sorry for delays reviews

PR #1068, opened 2023-02-15
Support startblock fetching for `graph add command`

incrypto32:


saihaj (approved):
Do we have a way today to add tests to existing suite?

incrypto32:
> Do we have a way today to add tests to existing suite?

Im unsure about that. Don't think there is

saihaj:
@incrypto32 can you rebase this and we can get this out

saihaj (approved):

PR #1193, opened 2023-03-17
fix: do not generate setters from derived fields

saihaj:
Derived fields are only meant for reverse lookups and not meant to be set, explained in [docs](https://thegraph.com/docs/en/developing/creating-a-subgraph/#reverse-lookups).

closes https://github.com/graphprotocol/graph-tooling/issues/943


dotansimha (approved):
LGTM.

PR #1195, opened 2023-03-19
Allow file data sources (with no network) to deploy to the studio

azf20:
Currently `graph deploy --studio ...` with a file data sources subgraph returns:
```
Error: The Subgraph Studio only allows subgraphs for these networks: mainnet, rinkeby, goerli, gnosis, chapel, optimism-goerli, clover, 
    fantom-testnet, arbitrum-goerli, fuji, celo-alfajores, mumbai, aurora-testnet, near-testnet, theta-testnet-001, osmo-test-4, base-testnet, celo, 
    arbitrum-one, avalanche
```
This is because all datasources for a subgraph are checked against that list, and file data sources don't have a network specified.
This PR skips that check if a data source doesn't have a network (at least one data source must have a network, so the deployment will be checked) - this works, but might not be a preferred approach

### DIFF
--- a/packages/cli/src/command-helpers/studio.test.ts
+++ b/packages/cli/src/command-helpers/studio.test.ts
@@ -33,6 +33,21 @@ describe('Version Command Helpers', () => {
         );
       });
 
+      test('And no network is passed', () => {
+        expect(() =>
+          validateStudioNetwork({
+            studio: true,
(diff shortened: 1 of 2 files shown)


schmidsi (approved):
No objections from my side for this approach.

PR #1197, opened 2023-03-20
fix: make generated list children non-nullable

saihaj:
This fixes the issue where the ABI -> GraphQL type was doing a simple 1:1 mapping where `Array<Bytes> -> [Bytes]` but for `graph-node` we require that list items are `non-nullable` so we need to make it `Array<Bytes> -> [Bytes!]`. I didn't want to introduce a big refactor so for now just making something that will work and accompanied test case. But [we should revisit refactoring this logic](https://github.com/graphprotocol/graph-tooling/issues/1198).

closes https://github.com/graphprotocol/graph-tooling/issues/1058

### DIFF
--- a/.changeset/small-foxes-fix.md
+++ b/.changeset/small-foxes-fix.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': patch
+---
+
+make generated list children non-nullable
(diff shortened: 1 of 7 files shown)


dotansimha (approved):
LGTM. @schmidsi @azf20 what do you think?

azf20 (approved):

enisdenjo (approved):

PR #1201, opened 2023-03-20
fix graph test command regressions

### LINKED ISSUE #1191: `graph test` command regressions

@graphprotocol/graph-cli

### Describe the issue

It seems there are some regressions in the `graph test` command since `v0.40.0`:

1. The `datasource` argument is made `required`, while it should be optional https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L16
2. `--version`, `-v` options is also marked as required, but it should be optional https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L51
3. It seems the `cachePath` will always point to the default location in the `./tests` folder https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L61-L62, which affects also the version cache https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L72. Specifying another test folder location through the `matchstick.yaml` file will do nothing, because the lookup happens at a later stage here https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L75 and it will always create and save the cache file in the default `./tests` folder instead in the one specified by the config.

### Reproduction

https://github.com/graphprotocol/graph-tooling/blob/main/packages/cli/src/commands/test.ts#L16

1. Run `graph test` without any arguments using graph-cli `0.42.4`

### Expected behavior

1. `graph test` should run without any arguments
2. First the lookup for a custom test folder from `matchstick.yaml` should happen before setting the `testDir` and `cachePath` and the the lookup for the cached matchstick version

### Screenshots or Videos
1. and 2.
Screenshot 2023-03-16 at 22 42 15

3.
Screenshot 2023-03-16 at 22 52 00

### Platform

- `@graphprotocol/graph-cli` "0.42.4"

### Subgraph Manifest

_No response_

### Subgraph GraphQL Schema

_No response_

### Additional context

saihaj:
Making `datasource` and `version` optional args to `graph test`

- [x] Need to find a way to repro and fix `cachePath` for `graph test`

Fixes #1191

### DIFF
--- a/packages/cli/src/commands/test.ts
+++ b/packages/cli/src/commands/test.ts
@@ -12,9 +12,7 @@ export default class TestCommand extends Command {
   static description = 'Runs rust binary for subgraph testing.';
 
   static args = {
-    datasource: Args.string({
-      required: true,
-    }),
+    datasource: Args.string(),
   };
 
   static flags = {
@@ -48,7 +46,6 @@ export default class TestCommand extends Command {
     version: Flags.string({
       summary: 'Choose the version of the rust binary that you want to be downloaded/used.',
       char: 'v',
-      required: true,
     }),
   };
 
@@ -135,7 +132,7 @@ async function runBinary(
     coverage: boolean;
     force: boolean;
     logs: boolean;
-    version: string;
+    version: string | undefined;
     latestVersion: string | null;
     recompile: boolean;
   },
@@ -253,7 +250,7 @@ async function runDocker(
     testsDir: string;
     coverage: boolean;
     force: boolean;
-    version: string;
+    version: string | undefined;
     latestVersion: string | null;
     recompile: boolean;
   },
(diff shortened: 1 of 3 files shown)


packages/cli/src/commands/test.ts
```diff
@@ -58,30 +55,17 @@ export default class TestCommand extends Command {
...
         // Load the config
         const config = await yaml.load(filesystem.read('matchstick.yaml', 'utf8')!);
-
-        // Check if matchstick.yaml and testsDir not null
-        if (config?.testsDir) {
+        // Check if matchstick.yaml and testsFolder not null
+        if (config?.testsFolder) {
           // assign test folder from matchstick.yaml if present
-          opts.testsDir = config.testsDir;
+          testsDir = config.testsFolder;
```
saihaj, line 68: we were not using the correct value here so it was just wrong

dimitrovmaksim (approved):
LGTM. Tested with the demo-subgraph in both binary and docker mode

enisdenjo (approved):

PR #1246, opened 2023-04-02
Fix add command overwriting abi

dimitrovmaksim:
fixes https://github.com/graphprotocol/graph-tooling/issues/1136

This PR should fix the bug where `graph add` renames the events in the ABI and manifest files, if there's an event name collision with an existing contract. This happens partially because `indexEvents` is `true` by default for the `add` command, which creates an entity in the `schema.graphql` for each event.

Now Instead of overwriting the `event._alias` with the prefixed event name, a `collision` property is added to the events which have collisions. When the manifest and the schema are scaffolded it will check for the `collision` property and will prefix the name only where it is needed, leaving the ABI and events signatures in the manifest file untouched.


packages/cli/tests/cli/add/subgraph/docker-compose.yml
```diff
@@ -0,0 +1,41 @@
+version: '3'
```
saihaj, line 1: can remove this

packages/cli/tests/cli/add/subgraph/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
saihaj, line 1: and this

packages/cli/tests/cli/add/subgraph/README.md
```diff
@@ -0,0 +1,4 @@
+# Example Subgraph
```
saihaj, line 1: this too

packages/cli/tests/cli/add/subgraph/networks.json
```diff
@@ -0,0 +1,7 @@
+{
+  "test": {
```
saihaj, line 2: same here

packages/cli/tests/cli/add/subgraph/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "basic-event-handlers",
+  "version": "0.1.0",
```
saihaj, line 3: and here

packages/cli/tests/cli/add/subgraph/src/mapping.ts
```diff
@@ -0,0 +1,21 @@
+import {
```
saihaj, line 1: don't think we need this either, since we are just asserting the `.json`

saihaj (dismissed):
Thank you! Change makes sense and thanks for test cases. Just nit to remove extra files form the example since we don't really need those for assertions. Once you remove those we can ship this.

packages/cli/tests/cli/add/subgraph/networks.json
dimitrovmaksim: This File is needed because the add commands updates the networks.json file

packages/cli/tests/cli/add/subgraph/src/mapping.ts
dimitrovmaksim: `codegen` after `add` fails if the mappings file does not exists

saihaj (approved):

saihaj (commented):
can you run `pnpm changeset` and add a patch for this. You can explain the fix there so it is documented and we can release. Learn more about changeset: https://github.com/graphprotocol/graph-tooling/pull/1246#issuecomment-1493416663 This will also give you a alpha release to try on your project.

saihaj (approved):

PR #1253, opened 2023-04-04
feat: disallow nullable primitive types

### LINKED ISSUE #1196: Codegen generated optional booleans has non-null assertion operator (!) by default which causes subgraphs to fail

@graphprotocol/graph-cli

### Describe the issue

consider this schema

```
type Assign @entity(immutable: true) {
  id: Bytes!
  to: Bytes! # address
  punkIndex: BigInt! # uint256
  blockNumber: BigInt!
  blockTimestamp: BigInt!
  transactionHash: Bytes!
  test: Boolean
  testRequired: Boolean!
  testNumberOptional: BigInt
}
```

The generated code for the getter for the `test` field in this Assign entity is 

```
  get test(): boolean {
    let value = this.get("test");
    return value!.toBoolean();
  }
```

it should not use the non-null assertion operator

instead it should handle the null value like how the testNumberOptional field handles it

```
  get testNumberOptional(): BigInt | null {
    let value = this.get("testNumberOptional");
    if (!value || value.kind == ValueKind.NULL) {
      return null;
    } else {
      return value.toBigInt();
    }
  }
  ```

### Reproduction

https://github.com/schmidsi/subgraph-schema-boolean-investigation

1. Run the matchstick test with `yarn test` after doing codegen and yarn install

it fails with 

```
  DESCRIBE ENTITY ASSERTIONS
thread 'main' panicked at '🆘 Unexpected error upon calling hook: Mapping aborted at generated/schema.ts, line 92, column 12, with message: unexpected null
wasm backtrace:
    0: 0x395f - <unknown>!start:tests/crypto-punks.test~anonymous|0~anonymous|1
', src/test_suite.rs:51:17
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error Command failed with exit code 101.

```

### Expected behavior

The test should run fine

### Screenshots or Videos

_No response_

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `@graphprotocol/*` version(s): [e.g. 2.6.2]

### Subgraph Manifest

_No response_

### Subgraph GraphQL Schema

_No response_

### Additional context

_No response_

### PR DESCRIPTION
**Pull request by saihaj**
In Assembly script `0,null, false` all will be represented the same way aka `0 byte` but the conversion we do today does not make sense. In GraphQL type system we can represent a nullable primitive type (i.e. a boolean or a int32) that when translated to TS we are asserting that it is not nullable which is false assumption because it will crash. With this change we are going to disallow nullable for primitive types.

closes #1196

### DIFF
--- a/packages/cli/src/codegen/schema.test.ts
+++ b/packages/cli/src/codegen/schema.test.ts
@@ -462,7 +462,9 @@ describe('Schema code generator', () => {
     try {
       codegen.generateTypes();
     } catch (err) {
-      expect(err.message).toBe(`A primitive type cannot be nullable. AssemblyScript does not support nullable primitives.\nConsider changing the type of "Account.isActive" from "Boolean" to "Boolean!"`);
+      expect(err.message).toBe(
+        `A primitive type cannot be nullable. AssemblyScript does not support nullable primitives.\nConsider changing the type of "Account.isActive" from "Boolean" to "Boolean!"`,
+      );
     }
   });
 });
--- a/packages/cli/src/codegen/schema.test.ts
+++ b/packages/cli/src/codegen/schema.test.ts
@@ -12,6 +12,7 @@ import {
   Param,
   StaticMethod,
 } from './typescript';
+import { writeFileSync } from 'fs-extra';
 
 const formatTS = (code: string) => prettier.format(code, { parser: 'typescript', semi: false });
 
@@ -31,6 +32,10 @@ const testEntity = (generatedTypes: any[], expectedEntity: any) => {
 
   for (const expectedMethod of expectedEntity.methods) {
     const method = methods.find((method: any) => method.name === expectedMethod.name);
+
+    if (!method) {
+      throw new Error(`Method ${expectedMethod.name} not found`);
+    }
 
     // eslint-disable-next-line @typescript-eslint/no-unused-expressions
     expectedMethod.static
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: packages/cli/src/codegen/schema.test.ts [hunk 1]
```diff
@@ -32,6 +32,10 @@ const testEntity = (generatedTypes: any[], expectedEntity: any) => {
   for (const expectedMethod of expectedEntity.methods) {
     const method = methods.find((method: any) => method.name === expectedMethod.name);
 
+    if (!method) {
+      throw new Error(`Method ${expectedMethod.name} not found`);
+    }
```
saihaj, line 37: makes life easier in test assertion in case fields are removed

packages/cli/src/codegen/schema.ts
```diff
@@ -295,19 +299,55 @@ Suggestion: add an '!' to the member type of the List, change from '[${baseType}
...
+  _typeFromGraphQl({
+    gqlType,
+    nullable = true,
+    entityDef,
+    fieldDef,
+  }: {
+    gqlType: immutable.Map<any, any>;
+    nullable?: boolean;
+    entityDef: any;
+    fieldDef: immutable.Map<any, any>;
```
saihaj, line 311: these are used so we can hint users what to change to make migration simpler : )

saihaj: I guess we can go one extra step and do `filename:location` this way user can click into error, kinda like how we ESlint or GraphQL codegen would do but that would require more refactor to pass the context of file around 🥲

saihaj:
superseded by https://github.com/graphprotocol/graph-tooling/pull/1284

PR #1258, opened 2023-04-04
feat: validate existence of specified handlers in manifest

### LINKED ISSUE #981: `graph build` does not check for existence of the handlers specified in the manifest
Build does not fail if there is an event / call / block handler whose function is not exported from the mapping file.

saihaj:
Today we do not run any validation when `graph build` runs. This PR introduces a way to load the generated `wasm` and ensure that it has all the handlers in it. If not we throw an error.

While working on this I did see and should be worked on in future
1. We have validation logic written but it is never invoked anywhere. We should figure out how to refactor the existing `Subgraph` class for the `Protocol` to provide better way to handle validation in a cheaper way.
2. Configuring unit tests for `graph build` are not straightforward since we are asserting the existence of `graph-ts` and  e2e just take longer to run  

closes #981

### DIFF
--- a/.changeset/silly-cougars-obey.md
+++ b/.changeset/silly-cougars-obey.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+validate existence for handlers specified in manifest during build
--- a/packages/cli/src/commands/build.ts
+++ b/packages/cli/src/commands/build.ts
@@ -106,7 +106,7 @@ export default class BuildCommand extends Command {
     if (watch) {
       await compiler.watchAndCompile();
     } else {
-      const result = await compiler.compile();
+      const result = await compiler.compile({ validate: false });
       if (result === false) {
         this.exit(1);
       }
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -289,7 +289,7 @@ $ graph create --node ${node} ${subgraphName}`;
         }
       });
     } else {
-      const result = await compiler.compile();
+      const result = await compiler.compile({ validate: true });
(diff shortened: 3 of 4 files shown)


dotansimha (dismissed):

dimitrovmaksim:
Just a thought, isn't it better to do the validation on deploy, оr at least have an option to skip it? I may want to check if my current code compiles correctly, before I have implemented all of the handlers. `init` (if I'm not mistaken) creates a handler only for the first event for example, which means I'll have to implement empty functions for each event I want to handle in order to run `build` and check if my current code compiles.

saihaj:
> Just a thought, isn't it better to do the validation on deploy, оr at least have an option to skip it?

given I can't run validation pre-bundling I think it is a good point to this check during deploy 🤔 @azf20 thoughts?

azf20:
thanks @dimitrovmaksim @saihaj as this is largely a failsafe I think it is reasonable to only check on deploy, particularly if it will slow down the build step

KholdStare:
@dimitrovmaksim @saihaj @azf20 Just to add some counter-points in support of adding validation during build:

- The point of the validation is to ensure the built WASM artifacts match the subgraph.yaml . If it lets some things through, like missing handlers, it's not doing its job. If there are missing handlers during development, then commenting out the events in `subgraph.yaml` temporarily is a better solution than crippling validation in general.
- Deployment is a last step, and usually not part of the quick development feedback loop. Seeing errors at that stage that could easily have been caught earlier is just frustrating. To add to that, this means event handler validation would be missing from CI pipelines, and "bad code" could be checked in that doesn't deploy.

I see that this PR adds this validation to the build step: https://github.com/graphprotocol/graph-tooling/pull/1545

PR #1277, opened 2023-04-11
fix: point to deployed API endpoint for IPFS

saihaj:
We need to append `/api/v0` if I just use `apiPath` as option for `create` then it removed `/ipfs` so better to point to the full URL

closes https://github.com/graphprotocol/graph-tooling/issues/1274


tomwhiteman:
> 

What do you mean? Please be specific.

PR #1306, opened 2023-04-25
feat: Add loadInBlock function for entities as part of codegen

### LINKED ISSUE #1305: Add loadInBlock function for entities as part of `codegen`
Provide a user friendly `loadInBlock(id)` function for entities to leverage the new `store.get_in_block` function created in https://github.com/graphprotocol/graph-node/pull/4540

i.e.
```
import { Transfer } from '../generated/schema'

let id = event.transaction.hash
let transfer = Transfer.loadInBlock(id)
```

If the entity was not created in the block, this function will return `null`

saihaj:
closes #1305


saihaj:
I did test on hosted service and is working as expected.

PR #1327, opened 2023-05-01
feat: do not generate any types for substreams

### LINKED ISSUE #1325: Update `graph codegen` for substream-based subgraphs
Currently running `graph codegen` on a subgraph with a substreams datasource creates a `generated` folder with types from the `schema.graphql`. If the `subgraph.yaml` only has a `substreams` datasource, no generated types are required, so `codegen` can return early (printing that information for the user)

saihaj:
**Before**

[asciicast](https://asciinema.org/a/LAQw3pQ27zgscuGM8pIOz3gZi)

**After**

[asciicast](https://asciinema.org/a/W0q2SWS4huvtFHdLedUcZ05ja)

closes #1325

### DIFF
--- a/.changeset/curvy-panthers-wink.md
+++ b/.changeset/curvy-panthers-wink.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+do not generate types, instead show a message to user to use substreams CLI for codegen.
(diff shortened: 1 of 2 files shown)


dotansimha (dismissed):

azf20:
Thanks @saihaj - I think the current text suggests that you need to do something with the Substreams CLI, where actually it's just an intentional no-op. Maybe:

✅ Subgraph uses a substreams datasource
✅ Code generation is not required

PR #1330, opened 2023-05-03
fix: all examples should install latest version

saihaj:
when someone clones the repo we do not want them to use `workspace:*` instead they should install whatever the `latest` tag is


enisdenjo (commented):
Just thinking about clones that might go stale. Like if I clone today, and run install N days after (containing changes) - the examples might not work because the deps pull the "latest".

Maybe still use concrete versions? Changesets should update them anyway so there's no maintenance.

dotansimha (approved):

enisdenjo (approved):

PR #1340, opened 2023-05-04
Derived Loader Fixes

incrypto32:
This PR completes https://github.com/graphprotocol/graph-tooling/pull/1190
- Fixes some issues with Immutable JS while rebasing 
- Fixes wrong type names being generated for derived loaders

### DIFF
--- a/packages/cli/src/codegen/schema.ts
+++ b/packages/cli/src/codegen/schema.ts
@@ -106,13 +106,18 @@ export default class SchemaCodeGenerator {
   }
 
   generateDerivedLoaders() {
-    const fields = this.schema.ast.definitions
-      .filter(def => this._isEntityTypeDefinition(def))
-      .flatMap((def: any) => def.fields)
-      .filter((def: any) => this._isDerivedField(def))
-      .map((def: FieldDefinitionNode) => this._getTypeNameForField(def.type));
+    const fields = (
+      (
+        this.schema.ast.definitions.filter(def =>
+          this._isEntityTypeDefinition(def),
+        ) as ObjectTypeDefinitionNode[]
+      )
+        .flatMap((def: ObjectTypeDefinitionNode) => def.fields)
+        .filter(def => this._isDerivedField(def))
+        .filter(def => def?.type !== undefined) as FieldDefinitionNode[]
+    ).map(def => this._getTypeNameForField(def.type));
 
-    return [...new Set(fields)].map((typeName: any) => {
+    return [...new Set(fields)].map(typeName => {
       return this._generateDerivedLoader(typeName);
     });
   }
--- a/packages/cli/src/codegen/schema.ts
+++ b/packages/cli/src/codegen/schema.ts
@@ -109,9 +109,10 @@ export default class SchemaCodeGenerator {
       .filter((def) => this._isEntityTypeDefinition(def))
       .flatMap((def: any) => def.fields)
       .filter((def: any) => this._isDerivedField(def))
-      .map((def: any) => this._getTypeNameForField(def));
+      .map((def: FieldDefinitionNode) => this._getTypeNameForField(def.type));
 
-    return [...new Set(fields)].map((typeName: any) => this._generateDerivedLoader(typeName));
+    return [...new Set(fields)].map((typeName: any) => {
+      return this._generateDerivedLoader(typeName)});
   }
 
   _isEntityTypeDefinition(def: DefinitionNode): def is ObjectTypeDefinitionNode {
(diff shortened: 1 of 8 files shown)


packages/ts/index.ts
```diff
@@ -26,6 +26,7 @@ export declare namespace store {
   /** If the entity was not created in the block, this function will return null. */
   // Matches the host function https://github.com/graphprotocol/graph-node/blob/9f4a1821146b18f6f49165305e9a8c0795120fad/runtime/wasm/src/module/mod.rs#L1091-L1099
   function get_in_block(entity: string, id: string): Entity | null;
+  function loadRelated(entity: string, id: string, field: string): Array<Entity>;
```
saihaj, line 29: can you add a changeset for explaining the upgrade to `graph-ts`

packages/cli/src/codegen/schema.ts
```diff
@@ -104,13 +105,31 @@ export default class SchemaCodeGenerator {
       .filter(Boolean) as Array<tsCodegen.Class>;
   }
 
+  generateDerivedLoaders() {
+    const fields = this.schema.ast.definitions
+      .filter(def => this._isEntityTypeDefinition(def))
+      .flatMap((def: any) => def.fields)
+      .filter((def: any) => this._isDerivedField(def))
+      .map((def: FieldDefinitionNode) => this._getTypeNameForField(def.type));
```
saihaj, line 113: let's not use `any` type here. When you filter ` .filter(def => this._isEntityTypeDefinition(def))` then it returns a `ObjectTypeDefinitionNode`  which is more narrowed down type

packages/cli/src/codegen/schema.ts
```diff
@@ -104,13 +105,31 @@ export default class SchemaCodeGenerator {
...
+
+    return [...new Set(fields)].map((typeName: any) => {
+      return this._generateDerivedLoader(typeName);
+    });
```
saihaj, line 117: same here let's avoid `any`'s

packages/cli/src/type-generator.ts
```diff
@@ -161,6 +161,7 @@ export default class TypeGenerator {
             GENERATED_FILE_NOTE,
             ...codeGenerator.generateModuleImports(),
             ...codeGenerator.generateTypes(),
+            ...codeGenerator.generateDerivedLoaders(),
```
saihaj, line 164: we should add changeset explaining the changes what this feature brings in, so we can have it documented in changelog.

saihaj (changes requested):
looks great in terms of feature. Few nits on code which we should fix and then good to merge

packages/cli/src/codegen/schema.ts
incrypto32: understood, but removing that any resulted in some weird errors, i have now manually casted these arrays to `ObjectTypeDefinitionNode[]` that seemed to have solved the problem

packages/cli/src/type-generator.ts
incrypto32: Added changesets

packages/cli/src/codegen/schema.ts
```diff
@@ -104,13 +105,36 @@ export default class SchemaCodeGenerator {
...
+  }
+
   _isEntityTypeDefinition(def: DefinitionNode): def is ObjectTypeDefinitionNode {
     return (
       def.kind === 'ObjectTypeDefinition' &&
       def.directives?.find(directive => directive.name.value === 'entity') !== undefined
     );
   }
 
+  _isDerivedField(field: any): boolean {
```
saihaj, line 132:
```suggestion
  _isDerivedField(field: DefinitionNode): boolean {
```

packages/cli/src/codegen/schema.ts
```diff
@@ -240,7 +326,56 @@ export default class SchemaCodeGenerator {
...
+      return this._returnTypeForDervied(gqlType.type);
+    }
+    if (gqlType.kind === 'ListType') {
+      return this._returnTypeForDervied(gqlType.type);
+    }
+    const type = tsCodegen.namedType(gqlType.name.value + 'Loader');
+    return type;
+  }
+
+  _generatedEntityDerivedFieldGetter(_entityDef: any, fieldDef: FieldDefinitionNode) {
```
saihaj, line 355: we can probably do `ObjectTypeDefinitionNode` here too 🤔

packages/cli/src/codegen/schema.ts
```diff
@@ -240,7 +326,56 @@ export default class SchemaCodeGenerator {
...
+      `get ${name}`,
+      [],
+      returnType,
+      `
+        return new ${returnType}('${entityName}', this.get('id')!.toString(), '${name}')
+      `,
+    );
+  }
+
+  _returnTypeForDervied(gqlType: any): any {
```
saihaj, line 344: need to define the return type

saihaj (approved):

PR #1382, opened 2023-06-20
fix: only ask for subgraph name and directory for init examples

saihaj:
### Summary

We don't need to know product or protocol information. The example comes with everything that is necessary. Even skipping the validation of networks since all the example that we have in this repo are maintained and should work.

[asciicast](https://asciinema.org/a/sufNmPJKbMnlDDunOKcCZ1SWa)

closes https://github.com/graphprotocol/graph-tooling/issues/1375

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -160,7 +160,8 @@ export default class InitCommand extends Command {
     const commands = {
       link: yarn ? 'yarn link @graphprotocol/graph-cli' : 'npm link @graphprotocol/graph-cli',
       install: yarn ? 'yarn' : 'npm install',
-      codegen: yarn ? 'yarn codegen' : 'npm run codegen',
+      // this ensure we run code-generation from our CLI instead of some script in package.json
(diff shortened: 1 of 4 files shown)


dotansimha (approved):

packages/cli/src/commands/init.ts
```diff
@@ -923,14 +971,14 @@ async function initSubgraphFromExample(
...
 
-  if (!isSubstreams) {
-    // Run code-generation
-    const codegen = await runCodegen(directory, commands.codegen);
-    if (codegen !== true) {
-      this.exit(1);
-      return;
-    }
-  }
+  // if (!isSubstreams) {
```
azf20, line 974: Did we intentionally remove the `codegen` step here?

azf20: Maybe that was removed as we are no longer asking for the `protocol`, but I think we can run `codegen` in all cases as it's a no-op for Substreams subgraphs?

saihaj: https://github.com/graphprotocol/graph-tooling/pull/1382/commits/4f03cf86d4d5b35fe775c1940189eaef3b0c9732

PR #1389, opened 2023-06-23
fix: returned string of BigInt.toHexString

fewwwww:
For The Graph's graph-ts [implementation](https://github.com/graphprotocol/graph-tooling/blob/main/packages/ts/common/numbers.ts#L113C1-L119C4), `BigInt.toHex()` and `BigInt.toHexString()` behaves the same. 

```js
toHex(): string {
  return typeConversion.bigIntToHex(this);
}

toHexString(): string {
  return typeConversion.bigIntToHex(this);
}
```

However, in my thoughts `toHex()` should return a string that looks like `deadbeef` while `toHexString()` should return a string that looks like `0xdeadbeef`.

So the implementation should be updated into:

```js
toHex(): string {
  return typeConversion.bigIntToHex(this);
}

toHexString(): string {
  return "0x" + typeConversion.bigIntToHex(this);
}
```

Additionally, `toHexString()` is not tracked by [The Graph's AssemblyScript API docs](https://thegraph.com/docs/en/developing/assemblyscript-api/#bigint).

### DIFF
--- a/packages/ts/common/numbers.ts
+++ b/packages/ts/common/numbers.ts
@@ -114,8 +114,15 @@ export class BigInt extends Uint8Array {
     return typeConversion.bigIntToHex(this);
   }
 
-  toHexString(): string {
-    return "0x" + typeConversion.bigIntToHex(this);
+  /**
+   * Converts `BigInt` to a hexadecimal string representation, with an optional
+   * prefix (for example, '0x').
+   */
+  toHexString(prefix: string = ''): string {
+    if (prefix !== '') {
+    return prefix + typeConversion.bigIntToHex(this)
+    }
+    return typeConversion.bigIntToHex(this);
   }
 
   toString(): string {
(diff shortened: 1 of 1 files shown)


packages/ts/common/numbers.ts
```diff
@@ -115,7 +115,7 @@ export class BigInt extends Uint8Array {
   }
 
   toHexString(): string {
-    return typeConversion.bigIntToHex(this);
+    return "0x" + typeConversion.bigIntToHex(this);
```
maoueh, line 118: This will be a major breaking change, while I kind of agree that we could have both I don't think we can merge that.

If AssemblyScript accepts default parameter, maybe we could think of `toHexString(format)` where default would be the same as today keeping compatibility and the format could be plain or prefixed or something along this.

maoueh (changes requested):
Will not merge in this form, and if AssemblyScript don't accept default parameter, we might even probably close this PR.

packages/ts/common/numbers.ts
fewwwww: Thanks for the review and feedback! I've updated the change with an optional `prefix` that does not affect the compatibility.

PR #1457, opened 2023-09-18
fix: loading interfaces for derived loaders

saihaj:
⚠️ TODO Need to explain ⚠️ 

closes https://github.com/graphprotocol/graph-tooling/issues/1449


dimitrovmaksim:
Does this change mean you can't do `parentEntity.childEntity.load()` on entities where `childEnitity` points to an interface? More precisely that `childEntity` field won't exists on the `parentEntity` at all?

dimitrovmaksim:
@saihaj Then the docs should point out that Interfaces are not supported for derived fields, because it seems some devs use them.

PR #1458, opened 2023-09-18
fix: do not create a git repo when already in a git repo

saihaj:
closes https://github.com/graphprotocol/graph-tooling/issues/1450

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -793,14 +793,14 @@ const initRepository = async (directory: string) =>
         filesystem.remove(gitDir);
       }
       if (await isInRepo()) {
-        await system.run('git init', { cwd: directory });
         await system.run('git add --all', { cwd: directory });
-        await system.run('git commit -m "Initial commit"', {
+        await system.run('git commit -m "Initialize subgraph"', {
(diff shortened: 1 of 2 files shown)


dotansimha (approved):

packages/cli/src/commands/init.ts
```diff
@@ -779,11 +792,18 @@ const initRepository = async (directory: string) =>
       if (filesystem.exists(gitDir)) {
         filesystem.remove(gitDir);
       }
-      await system.run('git init', { cwd: directory });
-      await system.run('git add --all', { cwd: directory });
-      await system.run('git commit -m "Initial commit"', {
-        cwd: directory,
-      });
+      if (await isInRepo()) {
+        await system.run('git init', { cwd: directory });
```
azf20, line 796: This looks like it's initialising a repository if it's in a repository, I think the opposite is the desired behaviour?

azf20:
Checking in on this @saihaj, did my comment make sense or did I read it wrong?

packages/cli/src/commands/init.ts
saihaj: a52e066dc38bea4d199ed10f75127088cd984841

PR #1462, opened 2023-09-21
fix: typos

omahs:
fix: typos

### DIFF
--- a/packages/cli/src/commands/local.ts
+++ b/packages/cli/src/commands/local.ts
@@ -1,94 +1,107 @@
-import { ChildProcess, spawn } from 'child_process';
-import http from 'http';
-import net from 'net';
-import path from 'path';
-import compose from 'docker-compose';
-import { filesystem, patching } from 'gluegun';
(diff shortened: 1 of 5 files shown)


packages/cli/src/commands/local.ts
```diff
@@ -50,7 +50,7 @@ export default class LocalCommand extends Command {
     'skip-wait-for-ipfs': Flags.boolean({
       summary: "Don't wait for IPFS to be up at localhost:15001",
     }),
-    'skip-wait-for-etherium': Flags.boolean({
```
saihaj, line 53: great catch but we would need to do a breaking change, we should deprecate this old arg, show a warning to user and drop this in a next major

omahs: @saihaj Thank you for your review. Unfortunately, I'm not a developer, and I wouldn't want to inadvertently cause any issues, so it would be greatly appreciated if you could handle the deprecation and warning tasks.

saihaj (approved):

PR #1470, opened 2023-09-27
feat: update test command to reflect new matchstick version

axiomatic-aardvark:


package.json
```diff
@@ -36,7 +36,8 @@
     "babel-jest": "^29.3.1",
     "eslint": "8.48.0",
     "jest": "29.7.0",
-    "prettier": "3.0.3"
+    "prettier": "3.0.3",
+    "semver": "^7.5.4"
```
saihaj, line 40: shouldn't be a global dep

saihaj (changes requested):

package.json
axiomatic-aardvark: my bad, still getting used to `pnpm` 😅

axiomatic-aardvark: @saihaj  should be good now

saihaj (approved):
thank you!

PR #1471, opened 2023-09-28
Updating `substreams` to latest version of crates

maoueh:
- Bumped `substreams` and `substreams-ethereum`
- Use `block.calls` instead of iterating through transactions
- Removed `buf.gen.yaml` that is not needed here
- Bumped to Rust 1.72 (why not while at it)


saihaj (approved):
thank you

PR #1489, opened 2023-11-01
feat: arbitrum sepolia

saihaj:


cmwhited (approved):
Think we need to wait to release. But looks good

PR #1491, opened 2023-11-03
Add retries to etherscan calls

### LINKED ISSUE #1447: increasing timeouts and more retries for external API calls
> Observation from the ETH Accra hackathon: Internet is not very stable and during init, there are some points where the CLI is doing some API calls which timed out: Download ABI from Etherscan and resolving start-block. Can we make these more resilient? Longer time-out windows and automatic retries?
from @schmidsi

YaroShkvorets:
Adds "Do you want to retry?" prompt when API calls fail during `graph init` wizard.
Fixes #1447

### DIFF
--- a/.changeset/honest-shoes-rhyme.md
+++ b/.changeset/honest-shoes-rhyme.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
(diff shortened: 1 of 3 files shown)


saihaj (approved):
looks great! thank you 🙏🏼

PR #1493, opened 2023-11-06
Fix codegen issues when using derived loader with Bytes as ID's

### LINKED ISSUE #1494: [Bug] Error loading derived entites for entity with `Bytes` ID
### Bug report

I'm facing an error when trying to [look up derived entities](https://thegraph.com/docs/en/developing/assemblyscript-api/#looking-up-derived-entities) using `.field.load()` in event handlers. This error happens when the Entity has the `id` field of type `Bytes`, but works fine when the id is of type `ID`(which is a string).
Tried debugging it and I think the issue happens in the getter for derived field. In my case the field name is `related`, so the getter looks like this: 
```
  get related(): RelatedBytesLoader {
    return new RelatedBytesLoader(
      "WithBytes",
      this.get("id")!.toString(),
      "related"
    );
  }
```
Looking at the error I'm guessing the compiler calls `toString` method of `Value` class which raises an assertion error(as seen in logs) because the type of id is `BYTES` not `STRING`. Possible fix could be to call `this.id.toString()`??
I've created a minimal reproducible example and attached details for the subgraph studio deployment.
Will add source code in a following comment

### Relevant log output

```Shell
Handler skipped due to execution failure, transaction: 0xc18a…71d8, address: 0xc944…44a7, signature: Transfer(indexed address,indexed address,uint256), error: transaction c18a251d1642abe44f048717345fe8f767fcbbd84c71a90653d9ce90002e71d8: Mapping aborted at ~lib/@graphprotocol/graph-ts/common/value.ts, line 80, column 5, with message: Value is not a string. wasm backtrace: 0: 0x253d - <unknown>!~lib/@graphprotocol/graph-ts/common/value/Value#toString 1: 0x2fc5 - <unknown>!src/handlers/handler
```

### IPFS hash

QmXWuYy7G69NsGnWjy6ybwZLrCVt43SqLUCaRgkEiKJTDY

### Subgraph name or link to explorer

https://thegraph.com/studio/subgraph/load-related-bytes-id-bug-4968

### Some information to help us out

- [ ] Tick this box if this bug is caused by a regression found in the latest release.
- [ ] Tick this box if this bug is specific to the hosted service.
- [X] I have searched the issue tracker to make sure this issue is not a duplicate.

### OS information

macOS

incrypto32:
Closes #1494

### DIFF
--- a/.changeset/witty-camels-switch.md
+++ b/.changeset/witty-camels-switch.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+Fix codegen issues when using derived loaders with Bytes as ID's
--- a/packages/cli/src/codegen/schema.test.ts
+++ b/packages/cli/src/codegen/schema.test.ts
@@ -537,6 +537,90 @@ describe('Schema code generator', () => {
     });
   });
 
+  test('get related method for WithBytes entity', () => {
+    const codegen = createSchemaCodeGen(`
+      type WithBytes @entity {
+        id: Bytes!
+        related: [RelatedBytes!]! @derivedFrom(field: "related")
+      }
+      
+      type RelatedBytes @entity {
+        id: ID!
+        related: WithBytes!
+      }
+    `);
+
+    const generatedTypes = codegen.generateTypes();
+
+    testEntity(generatedTypes, {
+      name: 'WithBytes',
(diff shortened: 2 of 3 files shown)


saihaj (approved):
nice tests. Thank you!

PR #1497, opened 2023-11-07
Etherscan rate limit

YaroShkvorets:
- Add rate limit parameter for some chains so `graph init` waits 5 seconds when needed
- Refactor chain definitions
- Replace dead etherscan-like api urls for some chains

### DIFF
--- a/packages/cli/src/command-helpers/chains.ts
+++ b/packages/cli/src/command-helpers/chains.ts
@@ -1,0 +1,139 @@
+interface NetworkInfo {
+  etherscanUrl: string;
+  publicRpcEndpoint: string;
+  rateLimit?: number;
+}
+
+export const networks: Record<string, NetworkInfo> = {
+  mainnet: {
+    etherscanUrl: 'https://api.etherscan.io/api',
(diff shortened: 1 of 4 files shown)


saihaj (approved):
thanks for the cleanup!

packages/cli/src/command-helpers/chains.ts
```diff
@@ -0,0 +1,139 @@
...
+  },
+  arbitrumOne: {
+    etherscanUrl: 'https://api.arbiscan.io/api',
+    publicRpcEndpoint: 'https://arb1.arbitrum.io/rpc',
+    rateLimit: 5000,
+  },
+  arbitrumGoerli: {
+    etherscanUrl: 'https://api-goerli.arbiscan.io/api',
+    publicRpcEndpoint: 'https://goerli-rollup.arbitrum.io/rpc',
+  },
```
saihaj, line 21: need to add arbitrum sepolia

YaroShkvorets: added

PR #1500, opened 2023-11-13
chore: upgrade to use node 20

saihaj:
- Drop node 16 from node test matrix
- include node 21 in node test matrix
- make node 20 the default env
- make node 20 the default dev env
- upgrade `pnpm` version
- CLI still will allow node 18+ just because it is still an active version from Node Releases WG

closes https://github.com/graphprotocol/graph-tooling/issues/1499

### DIFF
--- a/.changeset/fluffy-rockets-tickle.md
+++ b/.changeset/fluffy-rockets-tickle.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+upgrade to Node 20
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -19,7 +19,7 @@ jobs:
     with:
       npmTag: alpha
       buildScript: build
-      nodeVersion: 16
+      nodeVersion: 20
       packageManager: pnpm
     secrets:
       githubToken: ${{ secrets.GITHUB_TOKEN }}
@@ -31,7 +31,7 @@ jobs:
     with:
       npmTag: rc
       buildScript: build
-      nodeVersion: 16
+      nodeVersion: 20
       packageManager: pnpm
       restoreDeletedChangesets: true
     secrets:
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -4,7 +4,7 @@
   "description": "CLI for building for and deploying to The Graph",
   "license": "(Apache-2.0 OR MIT)",
   "engines": {
-    "node": ">=14"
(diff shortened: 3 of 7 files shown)


packages/cli/package.json
```diff
@@ -4,7 +4,7 @@
   "description": "CLI for building for and deploying to The Graph",
   "license": "(Apache-2.0 OR MIT)",
   "engines": {
-    "node": ">=14"
+    "node": ">=18"
```
enisdenjo, line 7: This would be considered a breaking change.

dotansimha: Agree with @enisdenjo 
Should release major in this case

.changeset/fluffy-rockets-tickle.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
```
dotansimha, line 2: should be `major` because of breaking changes

.changeset/fluffy-rockets-tickle.md
```diff
@@ -0,0 +1,5 @@
...
+---
+
+upgrade to Node 20
```
dotansimha, line 5: Is there anything needed from the consumer's point of view? Is anything important worth mentioning?

.github/workflows/pr.yml
```diff
@@ -31,7 +31,7 @@ jobs:
     with:
       npmTag: rc
       buildScript: build
-      nodeVersion: 16
+      nodeVersion: 20
```
dotansimha, line 34: Just a side note, I think at some point we can change this in `shared-config`; we can read the `node-version` file and just use it by default

dotansimha (approved):
LGTM. Just please verify the release type because of the breaking changes.

.github/workflows/pr.yml
enisdenjo: We can also use a .nvmrc file and point to it during setup-node with `node-version-file` option.

packages/cli/package.json
saihaj: we are still on `0.x` so it is a breaking change in that sense :)

.changeset/fluffy-rockets-tickle.md
saihaj: https://github.com/graphprotocol/graph-tooling/pull/1500#discussion_r1392713797

enisdenjo (approved):

PR #1501, opened 2023-11-14
upgrade prettier

saihaj:
this should fix the issue that time to time in vscode prettier is broken


enisdenjo (commented):
Why the `await`s suddenly in tests?

PR #1521, opened 2023-11-24
fix: remove validate studio network checks

### LINKED ISSUE #1509: Remove validateStudioNetwork during deploy
We are removing network limitations on the studio, so this check is no longer required.

saihaj:
Studio will take care of this and we don't need to maintain a list for supported networks.

closes #1509


packages/cli/src/commands/init.ts
```diff
@@ -245,8 +244,6 @@ export default class InitCommand extends Command {
           subgraphName,
           contractName,
           node,
-          studio,
```
azf20, line 248: Is this already removing some arguments from `init`?

azf20 (changes requested):
Nice! Does this already remove `studio` args from `init`? (Just checking if that was intended)

packages/cli/src/commands/init.ts
saihaj: no we are not removing any args. This is simply removing the validation check since we only do that check during `init` fro `studio`. Should not change any args.

azf20 (approved):

dotansimha (approved):

PR #1534, opened 2023-12-07
Add support for typecasting in mapping generation on cli init

travs:
This issue has been raised before (https://github.com/graphprotocol/graph-tooling/issues/949, https://github.com/graphprotocol/graph-tooling/issues/1017), but when generating a new subgraph for a contract that has events with `address[]` inputs, the generated mappings won't compile properly.

This is due to an issue where arrays of derived types (`Address[]` in this case) are not assignable to arrays of base types (`Bytes[]` here).

I wasn't sure how to fix this on a lower level as I'm not familiar with AssemblyScript, so I went and implemented a fix where the mapping assignments generated for events on `init` are conditionally wrapped with `changetype<typeB>(...parameter.of.typeA...)` if `typeA` is not able to be implicitly cast to `typeB`.

This got a bit messy and I had to extend some functions since we also must add imports for the types we are typecasting as (e.g. `Bytes`), and these must be calculated before we iterate events inside the template string.

I tested this locally and the generated mappings compile, but I'm not too sure about events with tuples, since they were a special case to begin with. A solution that doesn't do typecasting, but improves the base types or better uses AssemblyScript's existing type system is probably superior to my approach.

### DIFF
--- a/packages/cli/src/scaffold/index.ts
+++ b/packages/cli/src/scaffold/index.ts
@@ -153,7 +153,9 @@ dataSources:
 
   generateMappings() {
     return this.protocol.getMappingScaffold()
-      ? { [`${strings.kebabCase(this.contractName)}.ts`]: this.generateMapping() }
+      ? {
+          [`${strings.kebabCase(this.contractName)}.ts`]: this.generateMapping(),
+        }
       : undefined;
   }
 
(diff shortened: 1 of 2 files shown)


OKEAMAH (approved):

YaroShkvorets:
Hey I'm taking over this project from the previous maintainer. If you are interested in finishing up this PR please rebase on top of the latest `main`. If you could add a test or two and give an example of an existing contract that had this issue, that'd be great too.
Thanks.

YaroShkvorets:
Replaced by #1892

PR #1543, opened 2023-12-13
feat: integrate matchstick into the graph CLI

YassinEldeeb:
cleaning up / opening it for review in a couple of hours


packages/testing-matchstick/mocks/as/gravity-tests.txt
KholdStare, line 1: Dumb question - shouldn't this be `.ts` instead of `.txt`?

YassinEldeeb: It should be. but, we're currently just starting to move a direct clone of https://github.com/LimeChain/matchstick into here to maintain our own version. We haven't yet inspected the code fully 😄

Thanks for catching that! 😄

YassinEldeeb:
moved into #1549 where the branch lives on the repo for CI to run properly.

PR #1544, opened 2023-12-13
feat: validate graphql schema at build time

saihaj:
we want to run GraphQL Schema validations at the build time to crash early


OKEAMAH (approved):

OKEAMAH (approved):

YaroShkvorets:
@saihaj any reason this was abandoned as draft?

saihaj:
> @saihaj any reason this was abandoned as draft?

Hey @YaroShkvorets so if you see the `META_SCHEMA_GRAPH_NODE` there were some fix needed for schema in `graph-node` I don't think my PR ever got merged. Also this required some more testing. 

We also need a reliably way to pin to a `graph-node` version that a user will deploy to. There can be features from `graph-node` that user is trying to use in the subgraph but are not available in a release. So the approach of pinning to `master` on GitHub won't work. I think better would be looking at manifests and then creating URL for `META_SCHEMA_GRAPH_NODE`.

So that is one reason it was stranded. Then we also had bunch of discussions and exploration of solving this issue via the NAPI (or similar tool) initiative. Where `graph-node` would publish a common crate that can be used within CLI so we use same function `graph-node` uses and provide much more detailed error info. 

reflecting back I think better the former approach (the one this PR takes) is probably the best. Then in future figure out NAPI (we have PR that should be able to land after some extensive testing)

PR #1545, opened 2023-12-13
fix: validation for event handlers

saihaj:
Debugging an issue related to call handlers reported by @marcusrein and realized the changes I made last week are not needed because we already have a way to validate the WASM see: https://github.com/graphprotocol/graph-tooling/pull/1258

- Revert "fix: builds for templates (#1540)"
- Revert "feat: validation for handlers from manifest file (#1535)"
- fix: call handler validation

### DIFF
--- a/.changeset/@graphprotocol_graph-cli-1545-dependencies.md
+++ b/.changeset/@graphprotocol_graph-cli-1545-dependencies.md
@@ -1,0 +1,7 @@
+---
+"@graphprotocol/graph-cli": patch
+---
+dependencies updates:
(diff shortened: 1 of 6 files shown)


KholdStare:
Was about to file a bug report because `graph build` was crashing if a data source or template didn't have `eventHandlers`.

Just to double check - this will run the WASM validation during `graph build`?

saihaj:
> Was about to file a bug report because `graph build` was crashing if a data source or template didn't have `eventHandlers`.
> 
> Just to double check - this will run the WASM validation during `graph build`?

yep we are always running the `complie` for `WASM` in `build` so it just makes sense to enable validation here

PR #1554, opened 2023-12-19
test: more snapshots

saihaj:
instead of writing out big string by hand we should utilize more snapshots testing since that is what we are trying to achieve with these compares for generated code anyways.


saihaj:
I am not sure why I didn't merge this but goal was this makes test runner more stable.

YaroShkvorets:
Yeah I've re-implemented it in #1824

PR #1568, opened 2024-01-18
fix: zksync sepolia public rpc url

saihaj:
also added more debug logs to make it easier to debug


itsjerryokolo:
Screenshot 2024-01-18 at 3 47 27 PM
It works

PR #1623, opened 2024-04-08
fix: added Etherlink testnest ABI fetching API URL

Shiyasmohd:
Added Etherlink testnest ABI fetching API URL from Blockscout


saihaj (changes requested):
can you also add a Public RPC for this and run `yarn changeset` and create a `patch` for this change

saihaj:
Great work

.changeset/hip-spoons-promise.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
```
saihaj, line 2:
```suggestion
'@graphprotocol/graph-cli': patch
```

PR #1635, opened 2024-04-18
Add example subgraph for timeseries and aggregations

alex-pakalniskis:


examples/ethereum-cryptopunks-aggregations/schema.graphql
```diff
@@ -0,0 +1,18 @@
+type PunkBought @entity(timeseries: true) {
+  id: Int8!
+  timestamp: Timestamp!
+  punkIndex: BigInt!
+  value: BigInt!
+  fromAddress: Bytes!
+  toAddress: Bytes!
+  blockNumber: BigInt!
+  blockTimestamp: BigInt!
```
lutter, line 9: This isn't really needed - `timestamp` will also have the block timestamp

examples/ethereum-cryptopunks-aggregations/schema.graphql
```diff
@@ -0,0 +1,18 @@
+type PunkBought @entity(timeseries: true) {
+  id: Int8!
+  timestamp: Timestamp!
+  punkIndex: BigInt!
+  value: BigInt!
+  fromAddress: Bytes!
+  toAddress: Bytes!
+  blockNumber: BigInt!
```
lutter, line 8: Better to use `Int8` here

examples/ethereum-cryptopunks-aggregations/src/crypto-punks.ts
```diff
@@ -0,0 +1,16 @@
+import { PunkBought as PunkBoughtEvent } from '../generated/CryptoPunks/CryptoPunks';
+import { PunkBought } from '../generated/schema';
+
+export function handlePunkBought(event: PunkBoughtEvent): void {
+  let entity = new PunkBought('dummy-id');
```
lutter, line 5: For clarity, use `auto`; for any entity where the idtype is `Int8` or `Bytes`, you can set the `id` to `auto` and `graph-node` will make up a unique id. But we should really change `graph-cli` so that for time series, the generated code has a constructor that takes no arguments.

examples/ethereum-cryptopunks-aggregations/docker-compose.yml
```diff
@@ -0,0 +1,45 @@
...
+      POSTGRES_USER: graph-node
+      POSTGRES_PASSWORD: let-me-in
+      POSTGRES_DB: graph-node
+      # FIXME: remove this env. var. which we shouldn't need. Introduced by
+      # <https://github.com/graphprotocol/graph-node/pull/3511>, maybe as a
+      # workaround for https://github.com/docker/for-mac/issues/6270?
+      PGDATA: '/var/lib/postgresql/data'
+      POSTGRES_INITDB_ARGS: '-E UTF8 --locale=C'
+    volumes:
+      - ./data/postgres:/var/lib/postgresql/data
```
lutter, line 45: Do we want this in an example subgraph? I think it would be better to just have something in the `README` that points people at the official docs for how to set up a local environment

lutter (approved):
Looks great! I would like to also somehow mention or incorporate the [ethereum example](https://github.com/graphprotocol/common-subgraphs/tree/main/aggregation-example) as that shows more features of aggregations and explains a bit more how to use them.

Issue #323, opened 2019-08-09
On the init command, scaffold the subgraph with the contracts names of the ABIs


eboadom:
Currently the execution of `graph init` generates ABI's with the name **Contract** because of this https://github.com/graphprotocol/graph-cli/blob/3a9a5a972795029b63843761171fd7687e12fd8f/src/commands/init.js#L190 
Could be good to apply take as name the filename of the ABI instead of **Contract**, or even introduce another cli option on `graph init` to control that.

Jannis:
I'm not aware of any ABI files that contain the contract name, although there may be some. I agree a `--contract-name` or `--abi-name` CLI flag could work.

ernesto-usal:
I have no problem to submit a PR if you want @Jannis

Jannis:
@ernesto-usal That would be awesome! 😄

schmidsi:
We could add support for Hardhat (and others?) artifacts where the contract name and abi are in one file. Example of such a file is:
```json
{
  "_format": "hh-sol-artifact-1",
  "contractName": "MyNFT",
  "sourceName": "contracts/MyNFT.sol",
  "abi": [
    {
      "inputs": [],
      "stateMutability": "nonpayable",
      "type": "constructor"
    },
// ...
```

Lucifer0x17:
Is anyone working on the issue?

PR #1664, opened 2024-05-15
feat: auto fetch contract name from address

Shiyasmohd:
Auto fetches Contract name from contract address on `graph init` and `graph add` commands.
Solves #323


saihaj (approved):
great work!

Issue #1318, opened 2023-04-27
Autoupdate graph CLI


enisdenjo:
https://oclif.io/docs/releasing#autoupdater

PR #1749, opened 2024-10-25
Add a warning for available CLI updates

### LINKED ISSUE #978: Warn users on newest CLI version
We should notice CLI users about more recent versions of the CLI and encourage upgrades.

This can help on developers being both protected by possible security updates and be able to benefit from new features.

Implementation wise, that could be a message once per day instead of per command.

0237h:
Use [`@oclif/plugin-warn-if-update-available`](https://github.com/oclif/plugin-warn-if-update-available) to display a message to the user in case the CLI can be updated.

screenshot_2024-10-25_15-47-59

@alex-pakalniskis [Lots of options](https://github.com/oclif/plugin-warn-if-update-available?tab=readme-ov-file#configuration) for update checking frequency and user notification frequency.
Mainly what we might want to discuss:
- `timeoutInDays`: duration between update checks. Defaults to 60.
- `frequency`: the frequency that the new version warning should be shown (on every command execution by default).

Closes #978

We might want to also signal how to upgrade in the message. For now it could be a link to the release section of the repo or to documentation showing how to upgrade the CLI. Later on we could have a `graph update` command like #1318 is suggesting.

### DIFF
--- a/.changeset/wild-weeks-protect.md
+++ b/.changeset/wild-weeks-protect.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+Add warning for available CLI updates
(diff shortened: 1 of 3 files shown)


YaroShkvorets (approved):
LGTM

PR #1750, opened 2024-10-25
Add prompt for existing directory in `init` flow

### LINKED ISSUE #1370: Directory already exists when trying to initialize a subgraph and point it to an existing directory

@graphprotocol/graph-cli

### Describe the issue

When I initialize a new subgraph and select the current directory as "Directory to create the subgraph in", it fails with 
```
❯ Directory already exists
```

I think it is a valid use-case to create a subgraph in an existing directory. For example, when I need to store the ABI somewhere first, I create a folder to place the ABI. Now, I want to init the subgraph in this folder and not in another subfolder.

Maybe a confirm would do the trick? Like: "This directory already exists. Are you sure you want to initialize the subgraph there? This might have unforeseeable consequences"

### Reproduction

https://thegraph.com

graph init --studio base-token2
✔ Protocol · ethereum
✔ Subgraph slug · base-token2
? Directory to create the subgraph in › .  

### Expected behavior

Warning or just do it

### Screenshots or Videos

_No response_

### Platform

- OS: maxOX 13.2.1 (22D68)
- NodeJS: v18.15.0
- `@graphprotocol/*` version(s): @graphprotocol/graph-cli/0.50.1 darwin-arm64 node-v18.15.0

### Subgraph Manifest

_No response_

### Subgraph GraphQL Schema

_No response_

### Additional context

_No response_

0237h:
If the user is choosing an existing directory for creating a new subgraph, ask for confirmation.
If the answer is negative, abort the process.

Closes #1370

### DIFF
--- a/packages/cli/src/commands/init.ts
+++ b/packages/cli/src/commands/init.ts
@@ -634,12 +634,17 @@ async function processInitForm(
         name: 'directory',
         message: 'Directory to create the subgraph in',
         initial: () => initDirectory || getSubgraphBasename(subgraphName),
-        validate: value =>
-          filesystem.exists(value || initDirectory || getSubgraphBasename(subgraphName))
-            ? 'Directory already exists'
-            : true,
       },
     ]);
+
+    if (
+      filesystem.exists(directory) &&
+      !(await prompt.confirm(
+        'Directory already exists, do you want to initialize the subgraph here ?',
+        false,
+      ))
+    )
+      return;
 
     let choices = (await AVAILABLE_NETWORKS())?.[
       product === 'subgraph-studio' ? 'studio' : 'hostedService'


YaroShkvorets:
After confirming that I want to use the existing directory it proceeds but then fails with `Error: Directory or file "weth" already exists`. Should probably overwrite?

image

YaroShkvorets:
@0237h I think we should delete all files from the directory. Othewise we could be mixing different projects and/or committing a bunch of unrelated files into subgraph repository, which is not great.

Unless you can think of a scenario where we would want to keep any old files in the directory we are overwriting.

0237h:
I would argue that it could be more dangerous deleting files.

Right now, if you're not careful with the prompt, you'll end up overwriting/adding new files which isn't big deal. If we add deletion, one mistake and you could get files removed (any files).

Also there might be some use cases where you have ABIs in a subfolder that you might want to keep ?
Not entirely sure but in any case I would avoid deleting files and letting the user do it. Otherwise they should use a new directory, not existing one.

YaroShkvorets:
Yes, I agree about deleting being dangerous. Although I'm still having hard time picturing when it would be useful to keep existing files in your repo.
  
The ideal flow would probably be checking if directory exists right after the prompt and requiring to enter the different directory name. We can still keep "overwrite yes/no" prompt when generating the project. Can be a separate issue if that's too much work.

YaroShkvorets (approved):

PR #1760, opened 2024-11-06
update NETWORK_SUBGRAPH_MAINNET & add `api-key` to CLI

DenisCarriere:
fixes: https://github.com/graphprotocol/graph-tooling/issues/1753

## CLI changes
- add `--api-key`: The API key to use for the Subgraph queries.

**Example**
```
graph publish --subgraph-id 5zvR82QoaXYFyDEKLZ9t6v9adgnptxYpKpSbxtgVENFV --api-key <STUDIO API KEY>
```

## Website changes
- replace `NETWORK_SUBGRAPH_MAINNET` with https://gateway.thegraph.com/api/{api-key}/subgraphs/id/DZz4kDTdmzWLWsV373w2bSmoar3umKKH9y82SUKr5qmp
- replace `NETWORK_SUBGRAPH_SEPOLIA` with https://gateway.thegraph.com/api/{api-key}/subgraphs/id/3xQHhMudr1oh69ut36G2mbzpYmYxwqCeU6wwqyCDCnqV
- perform requests using decentralized network with Studio Subgraph API Key `?apiKey=<STUDIO API KEY>`
- Add notification in case `apiKey` is not provided in URL params

### DIFF
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -119,7 +119,13 @@ export default class PublishCommand extends Command {
     } = await this.parse(PublishCommand);
 
     if (ipfsHash) {
-      await this.publishWithBrowser({ ipfsHash, webapp: webUiUrl, subgraphId, protocolNetwork, apiKey });
+      await this.publishWithBrowser({
+        ipfsHash,
+        webapp: webUiUrl,
+        subgraphId,
+        protocolNetwork,
+        apiKey,
+      });
       return;
     }
 
--- a/packages/cli/src/commands/publish.ts
+++ b/packages/cli/src/commands/publish.ts
@@ -118,6 +118,13 @@ export default class PublishCommand extends Command {
       },
     } = await this.parse(PublishCommand);
 
+    if (subgraphId && !apiKey) {
+      ux.error(
+        'API key is required to publish to an existing subgraph (`--api-key`).\nSee https://thegraph.com/docs/en/deploying/subgraph-studio-faqs/#2-how-do-i-create-an-api-key',
+        { exit: 1 },
+      );
+    }
+
     if (ipfsHash) {
       await this.publishWithBrowser({
         ipfsHash,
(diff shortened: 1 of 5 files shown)


YaroShkvorets:
After looking into it I feel like the proper way would be to use `graph auth` to store the gateway API key. 
There already exists `graph auth` to store the deploy key for Studio. So introducing a separate flag just for `publish` seems wrong.

`graph auth` used to have `--studio` and `--hosted` flags. After removing Hosted service we removed both flags, so now it's only used storing Studio key

In Studio it actually tells you to using `--studio`, which doesn't work:

image

I suggest returning `--studio` flag to `graph auth` and also adding `--gateway` flag that would be used to store the gateway API key.

What do you think @alex-pakalniskis @0237h @DenisCarriere ?

0237h:
Good point, I think it makes sense to have all API keys managed from a single command like `auth`.

0237h:
> I propose returning --studio flag to graph auth and also adding --gateway flag that would be used to store the gateway API key.

@YaroShkvorets I'm wondering, aren't those technically the same keys ?
Is there a technical reason you need to use different keys or could users use the same for both and this is more of an option ?

@DenisCarriere qualifies the gateway key as just the *studio* API key and from what I understand it is also treated as such on the Explorer (see pic)
screenshot_2024-11-20_11-57-33

YaroShkvorets:
These are different keys.
The key to publish subgraphs is sent to `https://api.thegraph.com`. It's assigned to you by the Studio and it's not used for anything else.
The key to sent queries is sent to `https://gateway.thegraph.com` and you can manage/rotate it.

BTW looks like the website is still making some requests to the Network subgraph when you run `graph publish` with `--subgraph-id` flag to edit an existing one. Not sure where it's coming from. Need to fix that too.

image

YaroShkvorets:
1. make `--api-key` required for `graph publish`
2. figure out where network subgraph queries come from and replace with correct network subgraph url

YaroShkvorets:
Added version extraction from network subgraph for an existing subgraph to properly update metadata.

YaroShkvorets (approved):

Issue #1765, opened 2024-11-08
Handle non-existin files better


YaroShkvorets:
### Which packages are impacted by your issue?

@graphprotocol/graph-cli

### Describe the issue

If ABI/SPKG file is not accessible, the codegen just crashes.

<img width="1317" alt="image" src="https://github.com/user-attachments/assets/50adc99a-059b-41c7-babe-6f3ac0f367f4">

Should give an error and allow to re-enter.
Maybe even support TAB-based completion?

### Reproduction

n/a

### Steps to Reproduce the Bug or Issue

1. `graph init`
2. provide ABI from file
3. enter non-existing file name

### Expected behavior

See above

### Platform

- OS: [e.g. macOS, Windows, Linux]
- NodeJS: [e.g. 18.5.0]
- `@graphprotocol/*` version(s): [e.g. 2.6.2]

PR #1766, opened 2024-11-11
Improve ABI file path validation

0237h:
screenshot_2024-11-11_12-15-34

- Error message now shows within CLI interactive session instead of error crash
- Passing `--abi` will correctly pre-fill the interactive form
- Catch JSON parsing exception for `EthereumABI`.

Closes: #1765


YaroShkvorets (approved):
LGTM

works with SPKG files too

PR #1767, opened 2024-11-19
Fix CLI validation for `startBlock` and `contractName` fetched from external APIs

0237h:
If a value was retrieved via Etherscan-like APIs for `startBlock` and `contractName`, their value couldn't be overwritten by the user in the prompt.

This change corrects that behavior. The prompt can still be skipped by passing those values from the `init` command line directly (e.g. `--start-block=<value>`).


YaroShkvorets (approved):
LGTM

PR #1771, opened 2024-11-25
Support parsing full path contract name from block explorers

0237h:
Some explorers like [ZkSync](https://block-explorer-api.mainnet.zksync.io/docs#/Contract%20API/ApiController_getContractSourceCode) will return the `ContractName` with a full path like so:
```json
{
  ...
  "ContractName": "contracts/SyncSwapRouter.sol:SyncSwapRouter",
  ...
}
```
This PR adds support for parsing the contract name for this format (e.g. `SyncSwapRouter`).

### DIFF
--- a/.changeset/cyan-spoons-greet.md
+++ b/.changeset/cyan-spoons-greet.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': patch
+---
+
+Support parsing full path contract name from block explorers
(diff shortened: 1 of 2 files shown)


YaroShkvorets:
Looks like there are issues with openzeppelin contracts, i.e. `0xF77bd7c05598E094bc06e34bB81C07Bd3B091dB1`

image

YaroShkvorets (changes requested):
see comment

0237h:
@YaroShkvorets Good catch thanks, I've simplified the regex :+1:

0237h:
Simplified further without regex for cases like `yyy.sol:xxx` where the file name (`yyy`) doesn't match the actual contract class name (`xxx`).

Not sure if it's even possible but we don't know what the explorer can return so to be safe, this just takes whatever is after the last colon (`:`) character.

YaroShkvorets (approved):

PR #1778, opened 2024-11-29
fix: honor --contract-name and --start-block

briceyan:
This PR tries to address two issues related to subcommand add:

1. --contract-name flag ignored: When the --contract-name flag is specified in the command line, the provided value is ignored, and the user is prompted for input

2. Improved handling of --start-block flag: When the --start-block flag is provided, its value is correctly applied, but an unnecessary asynchronous call was still being made. This resulted in execution hanging due to an unhandled promise. The fix eliminates the redundant asynchronous call


0237h:
Hey @briceyan, thanks for the PR. Would you mind adding a description of the fix ?

I understand it fixes an issue with the `add` command not respecting flag parameters and I could reproduce the issue, appreciate the fix.

briceyan:
I initially suspected that the process hanging after completing the code generation was caused by some leftover async operations, but it now seems likely that prompt.ask is the reason for the hang. Unfortunately, I don’t have enough time to investigate this further at the moment—apologies for the incomplete diagnosis.

Below is the output when I ran the command with [why-is-node-running](https://www.npmjs.com/package/why-is-node-running):

------
...
+ why-is-node-running /Users/brice/Devel/graph-tooling/packages/cli/bin/run add --abi /Users/brice/EverythingFun/v1-core/out/ISet.sol/ISet.json --merge-entities --contract-name Set12 --start-block 1 0x0000000000000000000000000000000000000000
probing program /Users/brice/Devel/graph-tooling/packages/cli/bin/run
kill -SIGUSR1 19346 for logging
(node:19346) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
✖ Failed to fetch Contract Name: Failed to fetch contract source code: Contract source code not verified
✔ Contract Name · Set12
✔ Running codegen
There are 17 handle(s) keeping the process running.

# TTYWRAP
/Users/brice/Library/pnpm/global/5/.pnpm/why-is-node-running@3.2.1/node_modules/why-is-node-running/include.js:6 - console.log('kill -SIGUSR1', process.pid, 'for logging')

# TTYWRAP
(unknown stack trace)

# TTYWRAP
/Users/brice/Devel/graph-tooling/node_modules/.pnpm/password-prompt@1.1.3/node_modules/password-prompt/index.js:3 - const stdin = process.stdin

# FILEHANDLE
(unknown stack trace)

# FILEHANDLE
(unknown stack trace)
...

0237h:
@briceyan Not sure I fully understand, are you running into some other issue even after the fix of this PR ?

packages/cli/src/commands/add.ts
```diff
@@ -81,7 +81,7 @@ export default class AddCommand extends Command {
 
     if (isLocalHost) this.warn('`localhost` network detected, prompting user for inputs');
 
-    let startBlock = startBlockFlag;
+    let startBlock = parseInt(startBlockFlag ?? '').toString();
```
YaroShkvorets, line 84: if `startBlockFlag` is not provided this results in NaN and the whole command fails

packages/cli/src/commands/add.ts
```diff
@@ -160,8 +159,7 @@ export default class AddCommand extends Command {
 
     try {
       if (isLocalHost) throw Error; // Triggers user prompting without waiting for Etherscan lookup to fail
-
-      contractName = await loadContractNameForAddress(network, address);
+      contractName ||= await loadContractNameForAddress(network, address);
```
YaroShkvorets, line 162: While we are at it, could you move `contractName` uniqueness check from L87-L94 to below this try-catch so we can capture name loaded from etherscan too

packages/cli/tests/cli/add.test.ts
```diff
@@ -19,6 +19,8 @@ describe('Add command', () => {
         '0x2E645469f354BB4F5c8a05B3b30A929361cf77eC',
         '--contract-name',
         'Gravatar',
+        '--start-block',
```
YaroShkvorets, line 22: Add a separate test for case when startBlock is not supplied to avoid NaN bug above

YaroShkvorets (changes requested):
Thanks for the PR, it was indeed a bug. See my comments.

0237h (approved):

YaroShkvorets (approved):

PR #1782, opened 2024-12-06
cli: Add support for M4 apple silicon

prevostc:
* cli: Add support for M4 apple silicon


maoueh (approved):
We should probably switch to `/Apple (M[0-9])|processor)/` to overcome the ever updating list.

I leave that to current maintainer.

PR #1835, opened 2024-12-13
Fix generated example entity id uniqueness

YaroShkvorets:
Generated example entity id should be unique. See old PR #595


0237h (approved):
LGTM

PR #1888, opened 2024-12-30
Add support for proxy contracts

YaroShkvorets:
Implements #1739 

With `init` and `add` commands, if proxy is detected, the user is given an option to use ABI of the implementation contract.


.changeset/modern-goats-talk.md
```diff
@@ -0,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+Add EIP-1967/OpenZeppelin proxy contracts support. When proxy contract is detected, user is given an option to use ABI of the implementation contract.
```
0237h, line 5: Do we want to start only with those two proxy types in this PR, or should we expand to more ?

Found https://github.com/ahmedali8/detect-evm-proxy when researching, which we could use directly I guess although usage might not be trivial and it adds a dependency. But in any case it points out other types such as ERC-1167, etc.

packages/cli/src/commands/add.ts
```diff
@@ -100,6 +101,35 @@ export default class AddCommand extends Command {
           ),
         );
         if (!ethabi) throw Error;
+        const isProxy = ethabi.callFunctionSignatures()?.includes('upgradeTo(address)');
```
0237h, line 104: Not sure if that's the most reliable way to determine if contract could be a proxy.

We might catch false positives of contracts having this signature without being proxy (low chance I would guess but still). Maybe we could look for specific bytecode sequence or better heuristic ?

packages/cli/src/commands/init.ts
```diff
@@ -622,15 +623,49 @@ async function processInitForm(
...
+                    'Fetching implementation contract ABI...',
+                    'Failed to fetch implementation ABI',
+                    'Warning fetching implementation ABI',
+                    () =>
+                      contractService.getABI(protocolInstance.getABI(), network.id, implAddress!),
+                  ),
+                );
+              }
+            }
+          }
```
0237h, line 658: Probably need to make that part of the proxy detection function to avoid duplicating code here.

0237h (commented):
PR is in a good state for supporting main proxy types but some questions on extending the detection implementation and the proxy detection.

Also maybe use existing work like https://github.com/ahmedali8/detect-evm-proxy ?

.changeset/modern-goats-talk.md
YaroShkvorets: Didn't know there are more. Are they used though? Looks like EIP-1822 is stagnant: https://eips.ethereum.org/EIPS/eip-1822

Are there any examples that cannot be detected by the current implementation?

packages/cli/src/commands/add.ts
YaroShkvorets: This is the first check to avoid making unnecessary RPC calls. We then make an RPC call if `upgradeTo()` standard function is present in the ABI.

.changeset/modern-goats-talk.md
0237h: According to [this](https://dune.com/ilemi/erc-and-eip-starter-kit#contract-management) Dune dashboard, the three main proxy patterns are ERC1167 (Minimal Proxy), ERC1967 (covered right now), ERC2535 (Diamond pattern).

So I would say we're 1/3, would be best to cover 3/3 (you can check there's explanation at the top of the dashboard on how to track each).

packages/cli/src/commands/add.ts
0237h: Sure in that case if all the proxy implementation we aim to detect (see comment [above](https://github.com/graphprotocol/graph-tooling/pull/1888#discussion_r1907669307)) *all* implement this exact function, then that's a good check.

.changeset/modern-goats-talk.md
YaroShkvorets: Thanks for the dashboard link, very helpful.
Looks like "ERC2535 (Diamond pattern)" proxies to multiple contracts, so that would require some serious logic overhaul, so that's something for the future.
As for "ERC1167 (Minimal Proxy)", that appears to be an important one and would require reading bytecode. These proxies are not upgradeable and don't have `upgradeTo()` and `implementation()` functions.
I'll look into this one.

YaroShkvorets: Looked into ERC-1167, and it appears that since they are not upgradeable, when querying proxy ABI, Etherscan returns ABI of the implementation contract, so essentially these were always supported.

For example, `0x283e6a6E27365a7363a8E8011d5bd24735C8AE23` is a minimal proxy for `0x0ab72eb204001e24c2714f14cddfb8755ad9bc84` and Etherscan returns identical ABIs for them

So I think we are covered here.

packages/cli/src/commands/init.ts
YaroShkvorets: moved

.changeset/modern-goats-talk.md
0237h: Oh interesting they chose to do it like that. Agreed, should be fine then :+1:

0237h (approved):
LGTM, added test is awesome :fire:

Issue #1017, opened 2022-12-08
Address[] can not assign to Bytes[]


sdakhara:
I don't know what went wrong but when i try `npm run build`  which runs `graph build` it throws error mentioned bellow
```
Compile subgraphERROR TS2322: Type '~lib/array/Array<~lib/@graphprotocol/graph-ts/common/numbers/Address>' is not assignable to type '~lib/array/Array<~lib/@graphprotocol/graph-ts/common/collections/Bytes>'.

   entity.targets = event.params.targets;
                    ~~~~~~~~~~~~~~~~~~~~
 in src/dcode-dao.ts(43,20)

✖ Failed to compile subgraph: Failed to compile data source mapping: 1 compile error(s)
Error: Failed to compile data source mapping: 1 compile error(s)
    at Compiler._compileDataSourceMapping (/home/eu4/Desktop/DcodeGraph/node_modules/@graphprotocol/graph-cli/src/compiler/index.js:320:13)
    at /home/eu4/Desktop/DcodeGraph/node_modules/@graphprotocol/graph-cli/src/compiler/index.js:224:20
    at updateInDeepMap (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:1971:22)
    at updateInDeepMap (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:1980:23)
    at updateInDeepMap (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:1980:23)
    at Map.updateIn (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:1278:26)
    at /home/eu4/Desktop/DcodeGraph/node_modules/@graphprotocol/graph-cli/src/compiler/index.js:223:24
    at /home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:3016:46
    at List.__iterate (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:2206:13)
    at IndexedIterable.mappedSequence.__iterateUncached (/home/eu4/Desktop/DcodeGraph/node_modules/immutable/dist/immutable.js:3015:23)
UNCAUGHT EXCEPTION: Error: The AssemblyScript compiler crashed when compiling this file: 'src/dcode-dao.ts'
Suggestion: try to comment the whole file and uncomment it little by little while re-running the graph-cli until you isolate the line where the problem happens.
[...]

PR #1892, opened 2025-01-03
Add typecasting for tuple[] and address[]

### LINKED ISSUE #949: Error for array of address field.
With Array of address field `graph-cli` generate field as `[Bytes]` which fails while `codegen`.

On fixing `[Bytes]` to `[Bytes!]` in conversion file we are getting error below error as array of bytes cannot be directly assigned to array of address.

```
ERROR TS2322: Type '~lib/array/Array<~lib/@graphprotocol/graph-ts/common/numbers/Address>' is not assignable to type '~lib/array/Array<~lib/@graphprotocol/graph-ts/common/collections/Bytes>'.
```

Suggestion:
As we don't have support directly from `thegraph` for address type attributes (We get validation error), we need to generate code which will loop-over the array fields.

YaroShkvorets:
Fixes #949 #1017

### DIFF
--- a/.changeset/wild-pears-yell.md
+++ b/.changeset/wild-pears-yell.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+handle tuple[] and address[] for event parameters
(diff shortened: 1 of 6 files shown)


0237h (approved):
LGTM, kudos for test case :+1:

PR #1903, opened 2025-01-15
Fix `import.meta.url` not being parsed as path properly

### LINKED ISSUE #1881: `graph init`: whitespace in directory name
`yarn codegen` fails

```
✔ Install dependencies with yarn
✖ Failed to generate code from ABI and GraphQL schema: Command failed: yarn codegen
warning ../../../../package.json: No license field
node:fs:573
  return binding.open(

```

We should check directory name for invalid characters, including whitespaces.

0237h:
Using `import.meta.url` to resolve local file path and directories can lead to issues when special characters are percent-encoded in the url.

This fix uses the `fileURLToPath` native method to properly parse local urls to file paths.

Closes #1881

### DIFF
--- a/packages/cli/src/version.ts
+++ b/packages/cli/src/version.ts
@@ -7,7 +7,8 @@ const packageJson = JSON.parse(
     .readFileSync(
       // works even when bundled/built because the path to package.json is the same
       path.join(
-        `${process.platform === 'win32' ? '' : '/'}${/file:\/{2,3}(.+)\/[^/]/.exec(fileURLToPath(import.meta.url))![1]}`,
+        `${process.platform === 'win32' ? '' : '/'}${fileURLToPath(import.meta.url)}`,
+        '..',
         '..',
         'package.json',
       ),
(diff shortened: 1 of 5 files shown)


packages/cli/src/version.ts
```diff
@@ -1,12 +1,14 @@
...
 import path from 'node:path';
+import { fileURLToPath } from 'node:url';
 
 const packageJson = JSON.parse(
   fs
     .readFileSync(
       // works even when bundled/built because the path to package.json is the same
       path.join(
-        `${process.platform === 'win32' ? '' : '/'}${/file:\/{2,3}(.+)\/[^/]/.exec(import.meta.url)![1]}`,
+        `${process.platform === 'win32' ? '' : '/'}${fileURLToPath(import.meta.url)}`,
```
YaroShkvorets, line 10: Is this platform check needed any more? Can we test?

0237h: Could probably be removed as well yes according to a few sources:
- https://github.com/nodejs/node/issues/37845
- https://github.com/vitejs/vite/issues/17897

Although I don't have any windows to be able to test that.

YaroShkvorets (approved):
LGTM

PR #1929, opened 2025-01-28
`graph codegen`: handle reserved words

YaroShkvorets:
Fixes https://github.com/graphprotocol/graph-tooling/issues/1896

- Renames fields with reserved name in generated code "yield" -> "yield_"
- Handles all typescript/assemblyscript reserved words
- test


0237h (commented):
LGTM but just curious, why isn't `safeName` also used in the body of the generated code ?

For example here:
https://github.com/graphprotocol/graph-tooling/blob/f6ea8924b365aea798fbd645bad0e8a1e539c94a/packages/cli/src/codegen/schema.ts#L343-L352

YaroShkvorets:
That would mean renaming these fields in subgraph GraphQL schema, and that's unnecessary and leaks implementation details.

0237h (approved):

sins921 (commented):
- [ ]

PR #1934, opened 2025-01-31
fix: generate types for mappings when int8 is used as id

shiyasmohd:


YaroShkvorets:
Is this PR finished? 
`graph build` with these changes doesn't work if `id` type is set to `Int8`.
Codegen needs to be adjusted.

shiyasmohd:
@YaroShkvorets PR is finished now

YaroShkvorets:
@shiyasmohd 
Codegen now works, but scaffolded subgraph doesn't.

Steps to reproduct:
1. Init new subgraph: 
```
graph init test test --from-contract 0xdAC17F958D2ee523a2206206994597C13D831ec7 --protocol ethereum --network mainnet --index-events
```
2. Replace `id: Bytes!` for `Issue` entity in the GraphQL schema with `id: Int8!`
3. Run `graph codegen` - works
4. Run `graph build` - fails with 
```
 Compile subgraphERROR TS2322: Type '~lib/@graphprotocol/graph-ts/common/collections/Bytes' is not assignable to type 'i64'.

     event.transaction.hash.concatI32(event.logIndex.toI32())
     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 in src/contract.ts(30,5)

```

Basically need to generated
```

File: packages/cli/src/codegen/schema.ts [hunk 1]
```diff
@@ -16,6 +16,7 @@ import * as util from './util.js';
 class IdField {
   static BYTES = Symbol('Bytes');
   static STRING = Symbol('String');
+  static INT8 = Symbol('Int8');
 
   private kind: typeof IdField.BYTES | typeof IdField.STRING;
```
YaroShkvorets, line 21: add `| typeof IdField.INT8`

packages/cli/src/codegen/schema.ts
```diff
@@ -27,35 +28,99 @@ class IdField {
...
     const typeName = idField.type.type.name.value;
-    this.kind = typeName === 'Bytes' ? IdField.BYTES : IdField.STRING;
+    switch (typeName) {
+      case 'Bytes':
+        this.kind = IdField.BYTES;
+        break;
+      case 'Int8':
+        this.kind = IdField.INT8;
+        break;
+      default:
```
YaroShkvorets, line 38: let's add `case 'string'` for consistency and future refactoring

YaroShkvorets (changes requested):
One last thing, can we add a test case in `src/codegen/schema.test.ts`

shiyasmohd:
@YaroShkvorets PR is ready for review

PR #1977, opened 2025-02-28
Adding Subgraph Composition Repos

idalithb:
- Adding Subgraph Composition Repo Examples


YaroShkvorets (approved):
Changed links to point directly to repos to avoid error 404 when reading the readme.

PR #1984, opened 2025-03-09
fix: Fixed incorrect usage of `toI32()` for i64 conversion

mdqst:
I’ve fixed an issue where `toI32()` was being used to convert values to `i64`.
This is incorrect, as `toI32()` returns an `i32`, not an `i64`, which could result in data loss or unexpected behavior.

The issue has been resolved by updating the conversion to the correct function.


YaroShkvorets (approved):
Thanks for PR. Wasn't really a bug because underlying value is already an i32 since it's a 32-bit WASM runtime, but I agree using `toI64()` is better for consistency with return value

PR #1993, opened 2025-03-27
graph init should only generate entity handlers for immutable enitites

incrypto32:
We recently changed composed subgraphs to only work with immutable entities
This change in cli is focused on `graph init` to stop generating handlers for mutable entities

### DIFF
--- a/.changeset/short-coins-deny.md
+++ b/.changeset/short-coins-deny.md
@@ -1,6 +1,5 @@
 ---
 '@graphprotocol/graph-cli': minor
-'@graphprotocol/graph-ts': minor
 ---
 
 Composed subgraphs are modified to only accept immutable entites as triggers from a source subgraph
--- a/.changeset/short-coins-deny.md
+++ b/.changeset/short-coins-deny.md
@@ -1,0 +1,6 @@
+---
+'@graphprotocol/graph-cli': minor
+'@graphprotocol/graph-ts': minor
+---
+
+Composed subgraphs are modified to only accept immutable entites as triggers from a source subgraph
--- a/packages/cli/src/schema.test.ts
+++ b/packages/cli/src/schema.test.ts
@@ -1,4 +1,4 @@
-import { describe, expect, test } from 'vitest';
+import { beforeEach, describe, expect, test } from 'vitest';
 import Schema from './schema.js';
 
 describe('Schema', () => {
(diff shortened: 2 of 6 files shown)


YaroShkvorets (changes requested):
Looks good, missing a changeset.

packages/cli/src/schema.test.ts
```diff
@@ -0,0 +1,49 @@
...
+    expect(immutableEntityNames).not.toContain('Entity1');
+  });
+
+  test('getImmutableEntityNames handles explicitly non-immutable entities', async () => {
+    const schema = await Schema.loadFromString(schemaDocument);
+    const immutableEntityNames = schema.getImmutableEntityNames();
+    expect(immutableEntityNames).not.toContain('Entity3');
+  });
+
+  test('getImmutableEntityNames ignores non-entity types', async () => {
```
YaroShkvorets, line 43: This test essentially duplicates the previous two, and there are no non-entity types in the schema being tested.
Perhaps you meant to include non-entity types in the schema and test them here?

incrypto32: yeah i had a some non-entity types there initially then changed it but forgot to update the tests. Just updated to remove the test for now, we explicitly filter out for the entity directive any way.

.changeset/short-coins-deny.md
```diff
@@ -0,0 +1,6 @@
+---
+'@graphprotocol/graph-cli': minor
+'@graphprotocol/graph-ts': minor
+---
+
+Composed subgraphs are modified to only accept immutable entites as triggers from a source subgraph
```
YaroShkvorets, line 6: Looks like for `graph-ts` it removes `EntityTrigger` and `EntityOp` and this changeset message doesn't make much sense for it. Let's have separate changeset for graph-ts.

YaroShkvorets (changes requested):

.changeset/short-coins-deny.md
YaroShkvorets: I was thinking "Removed EntityTrigger and EntityOp types" to make it obvious this is a breaking change? But if you are saying it's not affecting users then either way is fine I guess, we don't have a to release it.

incrypto32: Just removed, and yes, this is not really a breaking change as this feature is not announced yet and there's no one using it currently.

YaroShkvorets (approved):

PR #1994, opened 2025-03-31
Require immutable flag on entities

isum:
This PR adds a new validation for GraphQL schemas that requires all entities to be explicitly marked as immutable or mutable. This is the first step in transitioning subgraph developers to using immutable entities by default.

As a result, the codegen command will fail if there are any entities without an immutable flag. The error message also provides a hint on what to update and how.

In addition, if GraphQL schemas do not use immutable entities _(all entities are mutable)_, the codegen command will print a warning suggesting that immutable entities should be considered and a link to best practices.

Finally, all example subgraphs have been updated with explicit immutable flags.

### Todos
- [x] Confirm the warning message with @alex-pakalniskis

### DIFF
--- a/.changeset/tangy-pants-buy.md
+++ b/.changeset/tangy-pants-buy.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+require immutable flag on entities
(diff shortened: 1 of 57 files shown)


matthewdarwin:
cosmos is no longer supported, we should consider removing those examples instead of updating them.

YaroShkvorets (approved):

PR #2015, opened 2025-05-05
refactor: log critical never

PaulRBerg:
Fixes https://github.com/graphprotocol/graph-tooling/discussions/2014


PaulRBerg:
I don't think `never` is supported in AssemblyScript so this PR might have to be reverted

PR #2047, opened 2025-06-06
update ipfs url

fubhy:


fordN:
Looks like we'll also need to update the tests at .`./packages/cli/src/command-helpers/compiler.test.ts`.

fubhy:
> Looks like we'll also need to update the tests at .`./packages/cli/src/command-helpers/compiler.test.ts`.

I reverted my changes to the test cases.

But after reading what's going on there I'm confused. This should not be a thing in the first place. Why would we magically rewrite some URLs and others not based on whether there's certain path segments in there or not.

That seems like a recipe for disaster / confusion. Just expect users to configure the right URL imo :-P

fubhy:
> Sorry out of the loop here. You are deprecating old E&N IPFS endpoints replacing them with `https://ipfs.thegraph.com/ipfs/{cid}`?
> 
> So you want the logic to be
> 
> * if the user supplies their own IPFS endpoint use that endpoint
> * if the supplied endpoint is one of those being deprecated print deprecation notice and use the new endpoint
> * if IPFS endpoint is not provided use the new endpoint (or make it required?)

Yeah. Exactly that. And no magic path substition logic (imo). That's a recipe for disaster.

Let's use the new one by default (not make it required).

YaroShkvorets:
Replaced by https://github.com/graphprotocol/graph-tooling/pull/2055

YaroShkvorets:
@fubhy opened #2055 if you want to review

PR #2083, opened 2025-10-26
feat: add network validation for subgraph contract addresse

### LINKED ISSUE #617: Check if subgraph contract addresses are in blockchain
If a user writes a subgraph referencing contract addresses from (for example) a testnet, and then accidentally deploys the subgraph to a graph node running mainnet, the subgraph will run but won't find any events. We can provide a better experience using [`eth_getTransactionCount`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactioncount) to check that the contract addresses exist on the blockchain in use.

I think it is technically possible to write a subgraph using contract addresses that are not yet present in the blockchain, but will eventually appear. Contract addresses are just a hash of the sender address and sender's nonce, so they could be determined ahead of time. This use case seems unusual however, and may not be something we want to support.

Open questions:
- error or just a warning?
- require at least one subgraph data source to have transaction(s) on chain, or all subgraph data sources to have transaction(s) on chain?

(thanks @leodasvacas for the suggestion)

Harshdev098:
fix: #617

Adds a pre-deployment check to warn users if subgraph contract addresses don’t exist on the target network.

### DIFF
--- a/.changeset/rare-falcons-cry.md
+++ b/.changeset/rare-falcons-cry.md
@@ -1,0 +1,5 @@
+---
+'@graphprotocol/graph-cli': minor
+---
+
+Network validation for subgraph contract addresses
--- a/packages/cli/src/commands/deploy.ts
+++ b/packages/cli/src/commands/deploy.ts
@@ -268,6 +268,32 @@ export default class DeployCommand extends Command {
       const dataSourcesAndTemplates = await DataSourcesExtractor.fromFilePath(manifest);
 
       protocol = Protocol.fromDataSources(dataSourcesAndTemplates);
+      for (const ds of dataSourcesAndTemplates.dataSources) {
+        const address = ds.source.address;
+        if (!address) continue;
+
+        try {
+          await client.request(
+            'eth_getTransactionCount',
+            [address, 'latest'],
+            (requestError, jsonRpcError, res) => {
+              if (jsonRpcError) {
+                const message = jsonRpcError?.message;
+                deployDebugger('message: %O', message);
+              } else if (requestError) {
+                deployDebugger('HTTP requestError: %O', requestError);
+              } else if (res === '0x0') {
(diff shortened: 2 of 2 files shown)


Harshdev098:
Hey @timmclean Can you please review it!

YaroShkvorets:
I'll have a look

YaroShkvorets (changes requested):
A couple of things:
1. sign the commits (this is a repo requirement): https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits
2. make it pass lint `pnpm lint` and tests `pnpm test`
3. add minor-version changeset `pnpm changeset`

Harshdev098:
@YaroShkvorets Thanks! Have done the required changes

YaroShkvorets:
Commit should have this "verified" badge.
image

Harshdev098:
But I have signedoff while amending the last commit @YaroShkvorets

YaroShkvorets:
You haven't added GPG key to your GitHub account to make it verified: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

Harshdev098:
@YaroShkvorets  Yah have done it!

YaroShkvorets (changes requested):
This approach doesn't work.

You are querying Graph node JSON RPC endpoint for transaction count when you should be querying EVM chain JSON RPC instead. Graph node doesn't support `eth_*` methods.

Technically you can load the networks registry, get JSON RPC endpoint for that chain and try querying it. Just keep in mind that there could be dead ones, non-evm chains, and other edge cases you need to handle.

Also before submitting a PR it's a good idea to verify the fix works at least on a happy path.

Harshdev098:
Hey @YaroShkvorets, Can you please review it! I have successfully tested it locally

PR #2125, opened 2026-05-14
feat(cli): delegate commands to gnd by default

lutter:
## Summary
  - `graph <cmd>` now execs `gnd <cmd>` (the Rust `@graphprotocol/gnd` binary) by default.
  - `graph local` stays on the TS path — gnd has no equivalent command.
  - New `graph dev` oclif command shims to `gnd dev` so the command stays discoverable via `graph --help` in TS mode.
  - Set `GRAPH_CLI_IGNORE_GND=1` to force every command through the oclif/TS implementation. The test suite sets this
  automatically in `tests/cli/globalSetup.ts`.

  Dispatch lives in `bin/run.js`; the spawn helper in `src/command-helpers/gnd.ts` is shared with `src/commands/dev.ts`.

  ## Test plan
  - [ ] `./bin/run.js --version` reports the gnd version
  - [ ] `GRAPH_CLI_IGNORE_GND=1 ./bin/run.js --version` reports the graph-cli version
  - [ ] `./bin/run.js local --help` shows the TS `local` help (with or without the env var)
  - [ ] `./bin/run.js dev --watch` actually starts gnd dev
  - [ ] `GRAPH_CLI_IGNORE_GND=1 ./bin/run.js dev --watch` also starts gnd dev (via the oclif shim)
  - [ ] `pnpm test:cli` passes (200/200)


incrypto32 (commented):
Few things i noticed testing this with my agent
1) `gnd add` doesn't append the new entities from the new contracts to the schema. So the generated subgraph fails to compile
2) `gnd init --protocol` near rejects any non 0x address. NEAR uses named accounts like wnear.flux-dev, not hex, so the validator blocks every valid input. gnd codegen also fails to parse `kind: near` manifests outright. NEAR subgraphs are broken on the new default.
3) `gnd codegen `can silently skip event signature validation. If subgraph.yaml references an event not in the ABI, gnd may still generate types and exit 0 with no warning, only to silently skip mappings at runtime.
4) Root cause for 3 is 
[load_abi](https://github.com/graphprotocol/graph-node/blob/4dca13398e1da81d316dba69eb6a87d093d69bd6/gnd/src/validation/mod.rs#L825) returns Option, silently drops parse errors. If alloy can't parse the ABI, signature validation silently skips . user sees "Types generated successfully" but no signatures were checked. Should return Result so failures surface.

